PR #48, opened 2026-05-07
fix(cli): walk Claude projects[] hierarchically for local scope

rstagi:
## Summary

- `ratel mcp import`/`link` now match how Claude itself resolves MCPs: discovery for the `local` scope walks `~/.claude.json`'s `projects[<dir>]` keys hierarchically from the current directory upwards (nearest-ancestor wins on collisions), instead of doing a single exact-key lookup at `projects[<projectRoot>]`.
- Stage B per-origin cleanup: `rewriteHomeClaude` removes migrated entries from each ancestor `projects[<X>]` they actually came from, and adds the `ratel` entry under `projects[<currentRoot>]` only.
- `defaultEnv()` captures `findProjectRoot` failures into `HierarchyEnv.projectRootError` instead of swallowing them; `import` and `link` log a one-line diagnostic listing the searched markers when no project root is detected.

## Why

Real-user repro: an early adopter's Claude Code `/mcp` showed 6 MCPs (slack, notion, perplexity, …) all stored under `projects["/Users/giacomo"].mcpServers`, but `ratel mcp import` from a sub-directory only surfaced the single root-level entry. Claude Code resolves descendant sessions to ancestor `projects[]` entries via inheritance; ratel's exact-key lookup didn't, so most users would lose entries on import depending on where they ran the wizard.

## Test plan

- [x] `pnpm -F @ratel-ai/cli test` (255 tests, +6 new covering parent-walk, nearest-wins, `localOriginByName`, missing-`projectRoot`, per-origin Stage B rewrite, and the diagnostic)
- [x] `pnpm -r typecheck` and `pnpm -r lint` clean across the workspace
- [x] Manual sandbox repro of the real-user case: from a no-marker sub-dir, all 7 ancestor entries surface (vs. previously only 1 root entry); diagnostic prints listing the searched markers and starting path
- [x] Stage B sanity: with a project marker present, ancestor `projects[<dir>].mcpServers` is emptied, `ratel` entry is added under `projects[<currentRoot>]` with the correct `--config` chain, and Ratel local config is written with all migrated entries


rstagi:
Closing as the premise of the issue turned out to be a fallacy

PR #54, opened 2026-05-13
TypeScript typehint for schema input outpus

### LINKED ISSUE #53: Feature request: support Standard JSON Schema in the TypeScript SDK
I tried the examples and hit this developer experience annoyance right away 😅

It would be nice to use libraries that support Standard JSON Schema https://standardschema.dev/json-schema in the TS SDK like Zod or Effect Schema, etc for tools input and output schema

kandros:
Add a barebone TS type hint for input/output schemas

~~I opted to use a minimal codebase-owned TS types definition instead of importing something like 
JSONSchema7 to avoid all the inconsistencies of different providers while improving DX~~

That became a can of worms faster than I expected, going with JSONSchema7 from @types/json-schema. 

There is a tradeoff to accept that not all providers support all features, but keeping that at the type level without specialized per-provider helpers is not worth it, overall it's a DX improvement

relates but don't resolve #53

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -92,6 +92,9 @@ importers:
       '@modelcontextprotocol/sdk':
         specifier: ^1.29.0
         version: 1.29.0(zod@4.4.1)
+      '@types/json-schema':
+        specifier: ^7.0.15
+        version: 7.0.15
     devDependencies:
       '@biomejs/biome':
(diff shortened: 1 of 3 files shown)


rstagi (approved):
Thank you @kandros ! Let me merge this one for now, then I'll think about the issue as a whole

PR #60, opened 2026-06-08
feat: Python SDK (ratel-ai) — full parity, PyO3-bound

kayraucklnc:
## Summary

Adds a **Python SDK (`ratel-ai`)** to the repo, at **full feature parity** with the TypeScript SDK (`@ratel-ai/sdk`). It binds the same Rust core (`ratel-ai-core`) via **PyO3 + maturin**, distributed as prebuilt `abi3` wheels — the exact Python analogue of the TS SDK's NAPI-RS binding. Python was on the v0.5.x roadmap; it landed early because the Rust core was already multi-language-ready (ADR-0009).

A Python agent can now `pip install ratel-ai` and use Ratel in-process, instead of the previous workaround (run the MCP server and connect to it).

## Architecture — mirrors the TS SDK exactly

Both SDKs are a thin two-layer wrapper over the **same** Rust core; no retrieval logic is reimplemented.

```
ratel-ai-core (Rust)  ──BM25, trace sinks──┐
            │ NAPI-RS                       │ PyO3
            ▼                               ▼
   ts/native (.node)                python/native (_native.abi3.so)   ← layer 1: pass-through
            ▼                               ▼
   catalog/gateway/mcp.ts   mirror   catalog/gateway/mcp.py            ← layer 2: ergonomics
     @ratel-ai/sdk (npm)             ratel-ai (PyPI)
```

API maps 1:1 (`searchToolsTool` → `search_tools_tool`, etc.), down to identical emitted trace events and **verbatim** gateway tool descriptions (a product contract shown to the model). Deliberate Python-idiomatic differences: `snake_case`, `asyncio`, and MCP ingestion as an optional extra (see Weak points).

## What's included

- **`src/sdk/python/native/`** — PyO3 binding over `ratel-ai-core` (`#[pyclass]`/`#[pymethods]` pass-through; `pythonize` preserves JSON-Schema key order per ADR-0004).
- **`src/sdk/python/ratel_ai/`** — pure-Python `ToolCatalog`, `search_tools_tool`/`invoke_tool_tool`, `register_mcp_server`, trace sinks; ships type stubs (`_native.pyi`, `py.typed`).
- **`docs/adr/0011-python-rust-binding-strategy.md`** — PyO3 decision (mirrors ADR-0002).
- **CI** — `python.yml` (PR gate: maturin + ruff + mypy + pytest), `python-binaries.yml` (5-target wheel matrix); `release.yml` extended (version-sync + CHANGELOG check + wheel build + **PyPI Trusted Publishing via OIDC**), `verify-install.yml` extended (pip smoke test).
- **`examples/pydantic-ai/`** — worked agent wiring (mirrors `examples/ai-sdk`).
- **Docs** — roadmap, AGENTS.md, README, CLAUDE.md, src/sdk README, lessons updated.

~840 lines of hand-written source + tests; the binary/FFI glue is generated by PyO3/maturin at compile time (nothing generated is committed).

## Verification (all green, run locally on macOS-arm64)

- **Rust**: `cargo build / clippy -D warnings / fmt / test --workspace` ✅
- **Python**: `ruff`, `mypy --strict`, `pytest` (31 tests) ✅
- **Packaging**: `maturin build --release` → abi3 wheel → installs into a **clean zero-dependency venv** → imports + searches ✅
- **Example**: diagnostic mode runs; tool-wiring validated against real **pydantic-ai 1.x** (`Tool.from_schema`, BM25 top-K + gateway) ✅

## Dry-run results (release workflow, run from this branch)

A `release.yml` dry-run from this branch **built all release artifacts green**:

- ✅ **All 5 Python wheels** — darwin-arm64, darwin-x64, linux-x64-gnu, linux-arm64-gnu, win32-x64-msvc.
- ✅ sdist, all 5 TS native binaries, version-sync + CHANGELOG checks.
- ⛔ The 3 publish jobs (crates/PyPI/npm) were correctly **rejected by the `release` environment's branch-protection** (a feature branch can't deploy to `release`). They run on a real tagged release from `main`. Not a failure — expected security behavior.

This **resolves the "wheel matrix unrun" risk below** — cross-platform Rust→Python builds are now proven on every target, including manylinux, linux-arm64 cross, and Windows.

## What's still needed before users can `pip install ratel-ai`

This PR ships the code and the publishing pipeline; it does **not** publish anything. Remaining steps (infra/release, not code):

1. **Cut a release** (version bumps — see next section).

## Releasing & version bumps

The release is **coordinated**: one `vX.Y.Z` tag re-publishes npm + cli + crate **and** the new Python wheel together, all at the same version (ADR-0008 version alignment, enforced by the `tag-version-check` job).

**Current state** — all four artifacts are at **0.1.5**, but **npm `@ratel-ai/sdk@0.1.5` is already published** (PyPI has nothing yet). Re-tagging `v0.1.5` would fail when npm/crate try to re-publish an existing version.

➡️ **So the first release that includes the Python SDK must bump to `0.1.6`** (or the next unreleased version). Python then gets its first publish at `0.1.6` alongside the others.

Steps to cut it:

1. Bump the version `0.1.5 → 0.1.6` in **all four** files (kept in lockstep, or the release aborts):
   - `Cargo.toml` (`version = "0.1.6"`)
   - `src/sdk/ts/package.json`
   - `src/integrations/cli/package.json`
   - `src/sdk/python/pyproject.toml`
2. Add a `## [0.1.6]` entry to **all four** CHANGELOGs (`src/core/lib`, `src/sdk/ts`, `src/sdk/python`, `src/integrations/cli`) — also enforced. The `/changelog` skill drafts these.
3. Merge that bump to `main`, then `git tag v0.1.6 && git push origin v0.1.6` → `release.yml` builds wheels and publishes (after the one-time PyPI Trusted Publisher setup above).

## How to review / try locally

```bash
# Rust core + binding
cargo build --workspace && cargo clippy --workspace --all-targets -- -D warnings

# Python SDK (needs uv)
cd src/sdk/python
uv venv --python 3.11 .venv
uv pip install --python .venv maturin pytest pytest-asyncio ruff mypy
.venv/bin/maturin develop
.venv/bin/ruff check . && .venv/bin/mypy ratel_ai && .venv/bin/pytest

# Example (diagnostic mode, no key needed)
cd ../../../examples/pydantic-ai && uv run main.py "read a config file and grep for TODOs"
```

Start the review at `src/sdk/python/native/src/lib.rs` (the binding) and `ratel_ai/{catalog,gateway,mcp}.py` (the surface), cross-referencing the TS originals under `src/sdk/ts/`.


examples/pydantic-ai/agent.py
```diff
@@ -0,0 +1,71 @@
...
+
+@dataclass
+class AgentResult:
+    text: str
+    active_tools: list[str]
+
+
+def _tool_from_executable(execute, name: str, description: str, schema: dict[str, Any]) -> Tool:
+    async def fn(**kwargs: Any) -> Any:
+        return await execute(kwargs)
```
rstagi: **[bug] Top-K tools crash: sync executor unconditionally awaited.**

`fn` does `return await execute(kwargs)`, but every catalog stub in `tools.py` is a *sync* lambda returning a dict (e.g. `tools.py:26`), so `await {dict}` raises `TypeError: object dict can't be used in 'await' expression`. Any tool selected by the BM25 top-K pre-filter — the example's headline path — therefore fails when the model calls it. The gateway tools work only because their `execute` is `async def`, and diagnostic mode never invokes a tool, which masks the defect.

The SDK itself handles this correctly in `ToolCatalog.invoke` via `inspect.isawaitable` (`src/sdk/python/ratel_ai/catalog.py:131`). Suggest mirroring that:

```python
async def fn(**kwargs):
    result = execute(kwargs)
    if inspect.isawaitable(result):
        result = await result
    return result
```

src/sdk/python/ratel_ai/mcp.py
```diff
@@ -0,0 +1,134 @@
...
+    return None
+
+
+async def register_mcp_server(
+    catalog: ToolCatalog,
+    *,
+    name: str,
+    session: Any,
+    transport_label: str = "unknown",
+    instructions: str | None = None,
```
rstagi, line 62: **[divergence] `server_instructions` / `transport` aren't auto-derived like the TS SDK.**

TS reads instructions from the live handshake (`client.getInstructions()`, `mcp.ts:26`) and derives the transport label from the transport's constructor name (`transportKind`, `mcp.ts:82`). Here both are manual params (`transport_label="unknown"`, `instructions=None`) echoed straight into the handle, so for the same upstream the handle/trace payloads differ from TS — `upstream_register.transport` is the literal `"unknown"` and `server_instructions` stays `None` unless the caller threads them through. That contradicts the module docstring's "the same `upstream_*` trace events the TS SDK emits."

I realise this partly stems from the deliberate ownership difference (the caller owns the `ClientSession` here), but the `initialize` result is available on the session, so instructions can still be auto-derived. Either derive them or drop the "identical events" framing.

.github/workflows/release.yml
```diff
@@ -42,12 +42,13 @@ jobs:
           ws_version=$(grep -E '^version' Cargo.toml | head -1 | sed -E 's/version *= *"([^"]+)"/\1/')
           sdk_version=$(node -p "require('./src/sdk/ts/package.json').version")
           cli_version=$(node -p "require('./src/integrations/cli/package.json').version")
+          py_version=$(grep -E '^version' src/sdk/python/pyproject.toml | head -1 | sed -E 's/version *= *"([^"]+)"/\1/')
```
rstagi, line 45: **[release wiring] `tag-version-check` will misbehave for RC tags.**

The repo is RC-first (`v0.1.5-rc.4`, …). This compares the raw `pyproject.toml` version string against the tag, but PyPI / PEP 440 normalizes `0.1.5-rc.4` → `0.1.5rc4`. So on an RC release the version published to PyPI differs from the npm/crate string, and to pass this check `pyproject.toml` must carry a non-normalized `0.1.5-rc.4` that won't match what actually lands on PyPI. Worth normalizing the comparison (or documenting the expected pyproject value for RCs).

.github/workflows/release.yml
```diff
@@ -246,9 +248,103 @@ jobs:
...
+          merge-multiple: true
+      - name: List artifacts
+        run: ls -la dist/
+      - name: Publish
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          packages-dir: dist
+          # Dry runs upload nothing; gate on the workflow_dispatch flag.
+          skip-existing: true
+        if: ${{ !inputs.dry_run }}
```
rstagi, line 357: **[release wiring] PyPI dry-run skips validation entirely.**

`if: ${{ !inputs.dry_run }}` skips the publish step on a dry run, whereas the npm and crate jobs run their publish with `--dry-run` (lines 198 / 248) and still validate packaging + metadata end-to-end. Consider running `gh-action-pypi-publish` against TestPyPI (or at least `twine check dist/*`) on dry runs so the PyPI path gets exercised before a real tag.

.github/workflows/verify-install.yml
```diff
@@ -74,3 +74,40 @@ jobs:
           npm init -y >/dev/null
           npm install --no-audit --no-fund "@ratel-ai/sdk@${{ steps.ver.outputs.spec }}"
           node -e "const {ToolRegistry} = require('@ratel-ai/sdk'); const r = new ToolRegistry(); r.register({id:'x',name:'x',description:'x',inputSchema:{},outputSchema:{}}); const hits = r.search('x', 5); if (!Array.isArray(hits) || hits.length === 0) { console.error('search failed'); process.exit(1); } console.log('OK', hits);"
+
+  verify-pypi:
+    name: verify-pypi (${{ matrix.label }})
```
rstagi, line 79: **[release wiring] `darwin-x64` wheel is published but never smoke-tested.**

`build-wheels` ships 5 wheels including `x86_64-apple-darwin`, and `verify-pypi` covers darwin-arm64 / linux-x64 / linux-arm64 / win32 — but not darwin-x64. The npm `verify` job deliberately covers darwin-x64 (under Rosetta, with an explanatory comment, ~line 31). A broken or mistagged darwin-x64 wheel would pass verification undetected. Suggest adding the matching darwin-x64 entry here.

src/sdk/python/CHANGELOG.md
```diff
@@ -0,0 +1,19 @@
+# Changelog
+
+All notable changes to `ratel-ai` (the Python SDK) are documented here.
+
+The format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) and this package adheres to [Semantic Versioning](https://semver.org/).
+
+## [Unreleased]
+
+## [0.1.5] - 2026-06-08
```
rstagi: **[doc/release] This documents a release that can't ship.**

`v0.1.5` was tagged 2026-05-10 and contains **zero** Python files, yet this dates a new `## [0.1.5] - 2026-06-08` "Initial release". Under Keep a Changelog a dated heading means "released on that date" — but 0.1.5 already shipped (without Python) a month earlier. The PR description itself concludes the first Python release must be `0.1.6`.

This entry should move under `[Unreleased]` (or become `[0.1.6]`), and `pyproject.toml:7` (`version = "0.1.5"`) bumped to match at release time. As committed the metadata describes a phantom version.

src/sdk/python/native/README.md
```diff
@@ -0,0 +1,23 @@
...
+Under the hood `maturin` compiles this crate (with the `extension-module` feature) into a platform-specific `abi3` extension importable as `ratel_ai._native`.
+
+## Layout
+
+```
+Cargo.toml      cdylib crate; depends on ratel-ai-core via the workspace
+src/lib.rs      #[pyclass] / #[pymethods] wrappers for ToolRegistry, SearchHit
+```
+
+The crate is a member of the top-level Cargo workspace, so `cargo build --workspace` picks it up automatically. `cargo build` / `cargo test` build *without* the `extension-module` feature (linking libpython so the tests run); `maturin` enables the feature for wheels.
```
rstagi: **[doc] This contradicts the crate's own `Cargo.toml`.**

It says `cargo test` builds without `extension-module` "so the tests run" — but `native/Cargo.toml` sets `test = false` (and `doctest = false`) and there are no Rust tests in this crate, so `cargo test` runs **zero** tests here. The Cargo.toml comment states the *opposite* rationale: the harness is disabled precisely so `cargo test --workspace` doesn't build a binary that links libpython and fails to find it at runtime. Suggest rewording to match what the build actually does.

src/sdk/python/README.md
```diff
@@ -0,0 +1,113 @@
...
+pip install ratel-ai
+# upstream MCP ingestion (register_mcp_server) needs the extra:
+pip install 'ratel-ai[mcp]'
+```
+
+Prebuilt `abi3` wheels ship for darwin-arm64, darwin-x64, linux-x64-gnu, linux-arm64-gnu, and win32-x64-msvc — no Rust toolchain required to install. The base SDK runs on CPython ≥ 3.9; the `mcp` extra requires ≥ 3.10.
+
+## Usage
+
+The SDK exposes two layers, both framework-neutral — the Python mirror of the [TypeScript SDK](../ts/README.md).
```
rstagi, line 28: **[doc] README drops the Telemetry section despite the "Python mirror of the TS SDK" claim.**

The TS README documents a `### Telemetry` subsection (the `trace` option and `noop` / `memory` / `jsonl` sink kinds — `src/sdk/ts/README.md:110-127`), and the Python SDK fully implements it (`TraceSinkConfig`, `ToolCatalog(trace=...)`, `drain_trace_events`). But this README has no trace / telemetry / jsonl mention, leaving a shipped, CHANGELOG-advertised feature undocumented.

examples/pydantic-ai/README.md
```diff
@@ -0,0 +1,45 @@
...
+```
+
+Splitting `tools.py` and `agent.py` keeps the catalog declarative and the loop readable; nothing about the wiring is provider-specific (`run_agent` accepts any Pydantic AI model id).
+
+## How the gateway works
+
+The agent's tool list at the start of the run is:
+
+- The **top-K** Ratel hits for the initial prompt — direct call, full schema visibility
+- **`search_tools(query, topK)`** — returns `[{toolId, score, description, inputSchema}, ...]`
```
rstagi: **[doc] `search_tools` return shape is mis-described (plus two smaller doc nits).**

- This line (and line 6) describe `search_tools` as returning `[{toolId, score, description, inputSchema}, ...]` / `{toolId, description, inputSchema}` triples, but the tool actually returns a **grouped** object: `{groups: [{server, hits: [...]}, ...]}` (`src/sdk/python/ratel_ai/gateway.py`).
- `src/sdk/python/README.md:93` references a "Cargo + (uv) workspace at the repo root" — there's no uv workspace at the root.
- `docs/roadmap.md` cites ADR-0009 (trace-events schema) as evidence the core was "multi-language-ready"; the binding-readiness ADRs are 0002 / 0011.

rstagi (commented):
Review of the Python SDK, focused on the parity / "identical" claims, the example, and release metadata. The core port (catalog/gateway/native) is faithful — verbatim gateway descriptions, matching `MAX_DESCRIPTION_LEN` and truncation logic, matching trace-event keys. The inline comments below are the issues worth addressing before merge/release; the first (broken example) and the version metadata are the blockers.

kayraucklnc:
Thanks @rstagi — all accurate, all fixed in 65b42c6.

**Bug**
- **agent.py async executor** — real break on the headline top-K path; now mirrors `ToolCatalog.invoke` (await only when `inspect.isawaitable`).

**Release wiring**
- **tag-version-check RC** — Python version now compared against the PEP 440 normalization (`0.1.5-rc.4` ↔ `0.1.5rc4`).
- **PyPI dry-run** — added `twine check dist/*` so packaging/metadata is validated on dry runs like npm/crate.
- **darwin-x64** — added to `verify-pypi` under Rosetta, matching the npm `verify` job.

**Docs**
- **CHANGELOG** — moved back under `[Unreleased]` (first Python release is the next bump, not the already-shipped 0.1.5).
- **native/README** — corrected: `test = false`, zero Rust tests; build links libpython so the workspace build/test passes.
- **python/README** — added the missing Telemetry section; dropped the non-existent root uv workspace claim.
- **examples README** — `search_tools` shape fixed to the grouped `{groups:[{server,hits:[...]}]}`.
- **roadmap** — binding-readiness now cites ADR-0002/0011, not ADR-0009.
- **mcp.py divergence** — took the 'drop the framing' option: scoped the claim to event *types*, and documented that `transport_label`/`instructions` are caller-supplied (pass the `initialize()` values for byte-identical `upstream_register`).

examples/pydantic-ai/agent.py
kayraucklnc: Turns out TS is okay with running await for sync functions it just ignores but Python/C etc will throw an error. Since I requested a mirror or TS we missed that also both in TS and Python we dont have end to end intergration tests we only had unit tests. There was a possibility to try to do: "await dict" which is okay in TS but not in python.

image

src/sdk/python/LICENSE.md
```diff
@@ -0,0 +1,41 @@
+Elastic License 2.0 (modified for open-source use)
```
rstagi, line 1: this actually needs to change. Let's merge it as it is, then please open a follow-up PR to update all the LICENSEs to MIT for now

rstagi (approved):
lgtm 🚀

PR #61, opened 2026-06-08
feat: first-class skills (BM25 retrieval, SDK catalog, tool-coupled suggestions)

kayraucklnc:
## TL;DR

Adds **first-class Skills** to Ratel — reusable Markdown playbooks, retrieved and served like tools instead of sitting always-on in the context window — in the shared **Rust core** and **both SDKs** (TypeScript + Python), shipping as the coordinated **`0.2.0`** release.

> ⚠️ **Breaking:** the discovery gateway tool is renamed `search_tools` → **`search_capabilities`** and now returns two reserved buckets `{ tools, skills }`. Hence `0.2.0`. The only consumer is `ratel-ai/ratel-mcp#7` (pre-release).

**Reviewer's 5-minute path:** `src/sdk/ts/src/gateway.ts` (the new surface) → `src/sdk/ts/src/skill-catalog.ts` + `skill-gateway.ts` → the parallel Python files (`src/sdk/python/ratel_ai/{gateway,skill_catalog,skill_gateway}.py`) → `src/core/lib/src/skill*.rs`. Everything else is wiring, tests, version, and docs.

---

## What this delivers

Skills are **on-demand playbooks**: an author writes a `SKILL.md` (frontmatter + body); the agent discovers it via search and pulls the body into context only when relevant — the same retrieval story Ratel already does for tools, applied to a second corpus. Three model-facing gateway tools:

- **`search_capabilities(query, topKTools?, topKSkills?)`** → `{ tools, skills }` — two **independently-ranked, separately-budgeted** buckets.
- **`invoke_tool(toolId, args)`** — run a tool.
- **`get_skill_content(skillId)`** — load a skill's body.

## Architecture & key decisions

- **Two reserved buckets, not one merged list.** Tools and skills are ranked against **separate BM25 corpora** with **independent top-K**. They are never merged or score-compared — BM25 scores from different corpora/term-statistics aren't comparable, and a flood of near-duplicate tool matches must not evict the one relevant skill. (`gateway.ts` `searchCapabilitiesTool`; corpora are `ToolRegistry` vs `SkillRegistry` in the core.)
- **`triggers` indexed, `stacks` not.** A `Skill` has `triggers` (author-declared task phrases — folded into the BM25 text so a terse intent prompt matches) and `stacks` (project stacks like `react` — carried for the *push-path* ranker in ratel-mcp to boost by context, deliberately **not** query terms). See `skill_indexing.rs`.
- **Both SDKs are thin mirrors of one core.** The skill logic lives in Rust (`SkillRegistry`, BM25). The TS SDK binds via NAPI, the Python SDK via PyO3 — each is a pure pass-through that constructs `core::Skill` and exposes a `SkillCatalog`. The Python binding is a line-for-line mirror of the TS NAPI (same methods, same param style). This is the ADR-0011 binding strategy.
- **Error contract.** Gateway error payloads now carry `isError: true` so a host (ratel-mcp's server) can promote it to MCP `isError` — the model can tell a failed call from real content. `invoke_tool` rejects a non-object `args` instead of forwarding stray top-level keys.

## What changed — by area

| Area | Change |
|---|---|
| **Rust core** (`src/core/lib`) | `Skill { …, triggers, stacks, body }`; a separate `SkillRegistry` BM25 index; `skill_indexing` (triggers in, stacks/body out); additive `TraceEvent` skill variants. |
| **TS SDK** (`src/sdk/ts`) | `gateway.ts` → `search_capabilities` + two buckets; new `skill-catalog.ts`, `skill-gateway.ts`; NAPI `Skill`/`SkillHit`/`SkillRegistry`; `index.ts` exports. |
| **Python SDK** (`src/sdk/python`) | **New parity:** PyO3 `SkillRegistry`/`SkillHit`; `skill_catalog.py`, `skill_gateway.py`; `gateway.py` → `search_capabilities` + buckets; `__init__.py`/`_native.pyi`. Same `isError`/args hardening. |
| **Release** | Lockstep `0.2.0` across `Cargo.toml`, `@ratel-ai/sdk` (+5 platform pkgs), `@ratel-ai/cli`, `ratel-ai`; four CHANGELOGs; merged with `main` (which carried the Python SDK + `0.1.6`). |
| **Docs/examples** | `search_tools→search_capabilities` swept across READMEs + the `pydantic-ai` example (its import was CI-run and would have broken); added a skills section to the Python README. |

## Breaking change & migration

`search_tools` → `search_capabilities`, result `{ groups }` → `{ tools: { groups }, skills }`.

```diff
- import { searchToolsTool } from "@ratel-ai/sdk"           // TS
- const r = await searchToolsTool(catalog).execute({ query, topK })
- r.groups
+ import { searchCapabilitiesTool } from "@ratel-ai/sdk"
+ const r = await searchCapabilitiesTool(catalog, skillCatalog?).execute({ query, topKTools, topKSkills })
+ r.tools.groups; r.skills
```
Python: `search_tools_tool` → `search_capabilities_tool` (same shape change). Only consumer is ratel-mcp#7.

## How it's verified — all three languages

- **Rust**: `cargo fmt --check`, `clippy --workspace -D warnings`, `cargo test --workspace`.
- **TS**: `pnpm -r build/typecheck/lint/test` — SDK **52** tests (incl. the two-bucket **no-starvation** test, `get_skill_content` body + `isError` error path + schema-shape guard), CLI **268**.
- **Python**: `maturin develop` → `ratel-ai-0.2.0`, `pytest` **41** (mirrors the TS tests), `ruff`, `mypy --strict`.
- **CI**: all `build` workflows (Rust/TS/Python) + the `example` (pydantic-ai) check pass on this PR.
- **End-to-end**: the unified surface is driven against a real MCP client + real `serve` in ratel-mcp#7 (`e2e/run.sh`, 17/17).

## Known limitations (deferred — not blocking, tracked)

- **Core BM25 ties are nondeterministic** — candidates collected in a `HashSet`, sorted with no secondary key; truncation before sort. Fix centrally in `bm25_search`: sort by `(score desc, id asc)`, truncate after. Affects the tools bucket too. _Recommended before GA._
- Data-safety items (skill-folder lifecycle) live in the consumer, tracked on ratel-mcp#7.

## Release sequence

Coordinated `0.2.0`: green CI → merge → tag `v0.2.0` → the release workflow publishes `@ratel-ai/sdk@0.2.0` (npm) + `ratel-ai==0.2.0` (PyPI) + `ratel-ai-core` (crates). Then ratel-mcp#7 bumps its SDK pin `^0.1.5 → 0.2.0`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/core/lib/src/search.rs
```diff
@@ -0,0 +1,64 @@
...
+        .map(|(id, contents)| Document { id, contents })
+        .peekable();
+    if documents.peek().is_none() {
+        return Vec::new();
+    }
+    let engine = SearchEngineBuilder::<String>::with_documents(Language::English, documents)
+        .k1(BM25_K1)
+        .b(BM25_B)
+        .build();
+    engine
```
rstagi: **Reintroduces the nondeterministic BM25 tie ordering that #63 just fixed on main.** This branch predates the fix, and the extracted `bm25_search` lets the engine cut to top-K before any stable ordering (the bm25 crate sorts by score only, from a HashSet), so tied scores flip on hash seed. Reproduced in smoke runs, in both buckets.

The merge conflict in `registry.rs` makes this worse: keeping this refactor silently drops the fix from the tool path, and main's two tie regression tests then fail flakily.

Fix in `bm25_search` itself: rank the full match set, sort by `(score desc, id asc)`, then truncate (sorting the returned top-K is not enough, the cut already happened). Add skill-side tie tests.

docs/adr/0011-first-class-skills.md
```diff
@@ -0,0 +1,82 @@
+# 11. First-class skills — unified capability search + two-path methodology
```
rstagi: ADR number collision: #60 already landed `0011-python-rust-binding-strategy.md` on main. Renumber this to 0012 (filename + heading); `src/sdk/python/native/src/lib.rs:3` references "ADR-0011" and now resolves to two documents.

src/sdk/ts/src/gateway.ts
```diff
@@ -213,14 +285,26 @@ export function invokeToolTool(
...
+          isError: true,
         };
       }
       const nested = inputObj.args;
-      const args =
-        nested && typeof nested === "object" && !Array.isArray(nested)
-          ? (nested as Record<string, unknown>)
-          : Object.fromEntries(Object.entries(inputObj).filter(([k]) => k !== "toolId"));
+      let args: Record<string, unknown>;
+      if (nested === undefined || nested === null) {
```
rstagi, line 314: An explicit `args: null` takes this fallback branch, and the filter only excludes `toolId`, so a stray `"args": null` key is forwarded to the tool. Same in Python (`gateway.py:282`).

src/sdk/ts/src/gateway.ts
```diff
@@ -253,7 +337,7 @@ export function invokeToolTool(
           tool_id: toolId,
           error: (err as Error).message ?? String(err),
         });
-        return { error: `tool ${toolId} threw: ${(err as Error).message}` };
+        return { error: `tool ${toolId} threw: ${(err as Error).message}`, isError: true };
```
rstagi, line 364: The `needs_auth` payload just above (line 328) is now the one error payload without `isError: true`, in both SDKs (`gateway.py:316`). Contradicts the CHANGELOG claim, and no test pins it.

src/sdk/ts/src/gateway.ts
```diff
@@ -184,14 +256,14 @@ export function invokeToolTool(
     name: INVOKE_TOOL_ID,
     description:
       "Invoke a tool from the catalog by its id. Use this to call tools that aren't in your direct tool list — " +
```
rstagi, line 277: This description and the unknown-`toolId` error text point the model to `search_capabilities`, which doesn't exist on a compat-only deployment using the deprecated `searchToolsTool`. Same strings in Python (`gateway.py:275`, `:333`).

src/sdk/python/ratel_ai/gateway.py
```diff
@@ -199,15 +272,26 @@ async def execute(input: dict[str, Any]) -> Any:
             )
             return {
                 "error": (
-                    f"unknown toolId: {tool_id}. "
-                    "Use search_tools to discover available ids."
-                )
+                    f"unknown toolId: {tool_id}. Use search_capabilities to discover available ids."
```
rstagi: Parity break: a call missing `toolId` raises `KeyError` (line 268; same for `skillId` in `skill_gateway.py:23`), while TS returns a structured `isError` payload. Recoverable error vs host crash on a model-facing surface.

docs/overview.md
```diff
@@ -65,15 +65,15 @@ This is the longer arc. Today's product is the foundation: a catalog, a retrieva
...
 
 ## How Ratel relates to MCP
 
 The Model Context Protocol is a transport / framing standard for tool-use between hosts and servers. Ratel is orthogonal: a retrieval engine that can sit on either side of an MCP boundary.
 
 - **As an MCP client**: `registerMcpServer` lets a Ratel catalog ingest the tools an upstream MCP server advertises, alongside any local executables. The agent in your process gets one unified catalog regardless of where each tool actually runs.
-- **As an MCP server**: `createMcpServer` wraps a catalog and exposes it over MCP — the connecting host (Claude Code, Cursor, ChatGPT) sees exactly two tools, `search_tools` and `invoke_tool`, regardless of catalog size. This is the path the `ratel` CLI uses to put Ratel between Claude Code and your existing MCP servers.
+- **As an MCP server**: `createMcpServer` wraps a catalog and exposes it over MCP — the connecting host (Claude Code, Cursor, ChatGPT) sees a small fixed set of tools (`search_capabilities` + `invoke_tool`, plus `get_skill_content` when skills are present), regardless of catalog size. This is the path the `ratel` CLI uses to put Ratel between Claude Code and your existing MCP servers.
```
rstagi, line 76: Claims the MCP server / CLI already expose `search_capabilities`; they still serve `search_tools` until ratel-ai/ratel-mcp#7 merges. Keep future-tense or land after the consumer.

rstagi (commented):
Reviewed in a worktree: all three stacks build and test green, and I smoke-tested the new surface in both SDKs (two-bucket search, `get_skill_content`, `invoke_tool` hardening, deprecated aliases). The feature works as advertised. Inline comments for the code issues; repo-level items below.

**Release / versioning**
- Versions and CHANGELOGs go straight to GA `0.2.0` (pre-dated 2026-06-10), contradicting our RC-first policy; an `-rc.1` tag would fail the release gates as-is. The `release.yml` changes themselves are fine (OIDC / Trusted Publishers respected).

**Docs drift**
- `AGENTS.md` and `llms.txt` still say skills "are on the roadmap"; root `README.md` and `docs/roadmap.md` likewise predate this PR.
- TS README documents none of the new skills surface; the Python README gained a full section in this same PR.
- `examples/ai-sdk/README.md` still says "v0.1.1 SDK" while using 0.2.0-only API; native binding folder READMEs weren't updated.

**Scope: this could have been 4 PRs.** Verified empirically by grafting each intermediate state onto main and running the full CI matrix: core, then TS SDK (+ CLI/examples), then Python SDK (+ example), then release prep. Every state is green. The core API is a strict superset (old natives compile unmodified), the rename never reaches core, natives pin core by path, and the tag-version-check blocks any mid-sequence publish, so lockstep constrains the tag, not merge order. A core-only first PR would also have surfaced the #63 conflict instead of leaving it latent. Happy to share the experiment details.

src/core/lib/src/search.rs
kayraucklnc: fixed

docs/adr/0011-first-class-skills.md
kayraucklnc: fixed, this was done before that one landed

src/sdk/ts/src/gateway.ts
kayraucklnc: fixed, stray args: null no longer forwards.

src/sdk/ts/src/gateway.ts
kayraucklnc: good idea to put isError in needs auth

src/sdk/ts/src/gateway.ts
kayraucklnc: would be a weird bug, fixed

src/sdk/python/ratel_ai/gateway.py
kayraucklnc: Recoverability is nice, getting Python into parity with TS too

docs/overview.md
kayraucklnc: Sure? Then we would re-open a PR just for this I vote to ship as is, we will merge MCP soon after

src/core/lib/src/tool_registry.rs
rstagi, line 1: maybe we should rename this to tool_registry to mirror the skill_registry?

src/core/lib/src/skill.rs
```diff
@@ -0,0 +1,18 @@
...
+/// are carried for the push-path ranker to *boost/filter* by project context
+/// rather than match as query terms. `body` is the dispatch payload and is also
+/// not indexed, so a long body never skews relevance.
+pub struct Skill {
+    pub id: String,
+    pub name: String,
+    pub description: String,
+    pub tags: Vec<String>,
+    pub triggers: Vec<String>,
+    pub stacks: Vec<String>,
```
rstagi: How are we using these exactly? This does not look like something that should live here, but rather on a higher level (e.g. the MCP).

The only things we care about for the skills are:
- id, name, description of course ✅
- body -> the instructions ✅
- tools -> the list of tools this skills might depend on ❌ -> we don't have this, but it's quite important to have a programmatic way to make an explicit dependency, and return those tools in the `search_capabilities` return value as well (we save multiple turns)
- tags -> optionally, we can add something like tags or metadata ❓ -> maybe we can use tags for "triggers" and "stacks"?

rstagi (dismissed):
Just last changes before merging, let's keep the models on Rust clean. Once fixed that feel free to merge it.

As we discussed in private:
- I would move the interface (all search/invoke tools) to Rust, and make the SDKs forward the calls to those one (only exception the invoke that will need an override to avoid encoding/decoding issues) -> not in scope for this PR, we'll tackle it in the next one

kayraucklnc:
Addressed the remaining review points on branch `claude/pr-61-review-feedback-djc57t` (stacked on this PR's head). Summary:

**1. `registry` → `tool_registry` ([rstagi]'s rename)** — module/file renamed to mirror `skill_registry` (`git mv`, history kept). The public type was already `ToolRegistry`; module is private, so no API change. Also fixed a `pub use` ordering that `cargo fmt --check` flagged. — `89ec2a9`

**2. Skill data model + explicit `tools` dependency** — the bigger one:
- Added `Skill.tools` — ids of tools a skill's instructions call. **Not** indexed (a dependency edge, not a ranking signal). `search_capabilities` now pulls a matched skill's declared tools into the `tools` bucket, **additively** (beyond `topKTools`) and **deduped** against query hits, so the agent gets the playbook + its toolkit in one turn instead of a second search. — `1ef2f1c`
- **Collapsed `triggers`/`stacks`** per your comment: `triggers` fold into `tags` (both were just indexed phrases); `stacks` move under a generic, non-indexed `metadata: map<string, string[]>` (`metadata["stacks"]`) carried for the push ranker. Final shape: `Skill { id, name, description, tags, tools, metadata, body }`. This keeps ADR-0012's "context is a boost, not a query term" invariant (tags indexed, metadata not). — `ae37585`
- ADR-0012, READMEs, CHANGELOGs updated to the new model (ADR edited in place since it hasn't merged). — `ae37585`, `5f8e192`

Threaded through core + NAPI + PyO3 + both SDK types, with parity tests on both sides. Green across all three: Rust (fmt/clippy `-D warnings`/test), TS (60 SDK + 268 CLI), Python (ruff/mypy/41→51 pytest).

**3. Moving the whole search/invoke interface into Rust (SDKs forward)** — kept out of scope as agreed; tracked as a follow-up.

---
_Generated by [Claude Code](https://claude.ai/code)_

rstagi:
@kayraucklnc did you forget to push some commits?

kayraucklnc:
I had a couple questions in mind I wanted to check, didnt wanted to merge, today will do @rstagi

Issue #56, opened 2026-05-20
Ambiguous tool-ranking case: documentation query ranks unrelated weather tool above docs tool


k33wee:
## Summary
I found a small reproducible ambiguous tool-ranking case while testing `@ratel-ai/sdk` locally with a toy tool catalog.
For a documentation/configuration-related query, the expected tool `search_docs` is retrieved in the top-K, but an unrelated tool, `weather_lookup`, ranks above it.
This seems useful as a diagnostic case because it looks like a ranking-quality issue rather than a recall issue.

Edited: Updated script with correct reproduction step

## Reproduction
```ts
import { ToolCatalog } from "@ratel-ai/sdk";

const catalog = new ToolCatalog();

const makeTextTool = (
  id: string,
  description: string,
  handler: (args: Record<string, unknown>) => Promise<Record<string, unknown>>,
) => {
  catalog.register({
    id,
    name: id,
    description,
    inputSchema: {
      properties: {
        query: {
          type: "string",
          description: "Natural language request or search query",
        },
        path: {
          type: "string",
          description: "Optional file path or directory path",
        },
        pattern: {
          type: "string",
          description: "Optional keyword or regular expression pattern",
        },
      },
    },
    outputSchema: {
      properties: {
        summary: {
          type: "string",
          description: "Textual summary of the tool result",
        },
      },
    },
    execute: handler,
  });
};

makeTextTool(
  "read_file",
  "Read a file from local disk and return its textual contents.",
  async ({ path }) => ({
    summary: `Read file at ${String(path ?? "<missing>")}`,
  }),
);

makeTextTool(
  "search_code",
  "Search source code across a repository for keywords, symbols, or configuration values.",
  async ({ pattern }) => ({
    summary: `Searched code for pattern ${String(pattern ?? "<missing>")}`,
  }),
);

makeTextTool(
  "search_docs",
  "Search project documentation, markdown files, and technical docs for relevant explanations.",
  async ({ query }) => ({
    summary: `Searched docs for ${String(query ?? "<missing>")}`,
  }),
);

makeTextTool(
  "search_logs",
  "Search application logs and runtime logs for errors, warnings, and diagnostics.",
  async ({ query }) => ({
    summary: `Searched logs for ${String(query ?? "<missing>")}`,
  }),
);

makeTextTool(
  "read_env_file",
  "Read environment variable files such as .env, .env.production, or deployment config files.",
  async ({ path }) => ({
[...]

PR #63, opened 2026-06-10
fix(core): deterministic ordering for tied BM25 scores in tool search

rinaldofesta:
## Summary

Found while reproducing #56: exact BM25 score ties are ordered nondeterministically across process runs. In k33wee's 12-tool catalog, scenario "read the production environment variables file" puts `send_email` and `weather_lookup` at exactly `0.03990466520190239`, and which one makes the top-5 cut flips between runs (5/5 over 10 Bun runs, 4/4 over 8 Node runs in my samples). Top-K membership flickers on ties, and small-catalog repros become easy to mismatch.

The cause is upstream of the registry: the bm25 crate collects candidate documents through a `HashSet` and sorts by score only, so equal scores keep hash-seed iteration order. This does not touch the ranking inversion reported in #56 (those scores are not tied); diagnosis and a proposed direction for that half are in the issue thread.

## Change

- `search_with_origin` ranks the full match set, breaks ties by `tool_id`, then cuts at `top_k` (`src/core/lib/src/registry.rs`).
- Two integration tests in `tests/tool_search.rs`: tied scores ordered by `tool_id`, and stable top-K membership with a tie at the boundary. Pre-fix they fail on hash-seed order: across 12 fresh `cargo test` processes I got 8 runs with one failing and 4 with both failing.
- No API change. Scores and non-tie orderings are unchanged: the original #56 repro output against a locally built SDK is byte-identical to a pre-fix run, and hashes identically across 10 consecutive runs (pre-fix: two variants, 5/5).

Note on cost: the engine already rebuilds per query over the whole catalog, so ranking the full match set instead of `top_k` adds a sort over the matched subset, nothing asymptotically new. If you would rather keep the early cut and only re-sort the returned page, that loses cross-page determinism at the boundary; full-set ranking is the smaller invariant to reason about.

## Test

- `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` (31 passed) all clean.
- TS SDK with the rebuilt native module: `pnpm test` 51 passed, `pnpm lint`, `pnpm typecheck` clean.
- Not run: the Python SDK test suite and ratel-bench campaigns.

#61 touches `registry.rs` too; happy to rebase this after it lands if that ordering is easier.

---
*Implemented with the help of a coding agent (Claude Code) and reviewed locally by the author.*

### DIFF
--- a/src/core/lib/src/registry.rs
+++ b/src/core/lib/src/registry.rs
@@ -78,14 +78,27 @@ impl ToolRegistry {
                 .k1(BM25_K1)
                 .b(BM25_B)
                 .build();
-            engine
-                .search(query, top_k)
+            // The bm25 crate sorts by score only and collects candidates through
+            // a HashSet, so equal scores fall back to hash-seed iteration order
+            // and flip between processes: tied hits at the top_k boundary make
+            // top-K membership nondeterministic. Rank the full match set, break
+            // ties by tool_id, then cut.
+            let mut ranked: Vec<SearchHit> = engine
+                .search(query, self.tools.len())
                 .into_iter()
                 .map(|r| SearchHit {
                     tool_id: r.document.id,
                     score: r.score,
                 })
-                .collect()
+                .collect();
+            ranked.sort_by(|a, b| {
+                b.score
+                    .partial_cmp(&a.score)
+                    .unwrap_or(std::cmp::Ordering::Equal)
+                    .then_with(|| a.tool_id.cmp(&b.tool_id))
+            });
+            ranked.truncate(top_k);
+            ranked
         };
         let took_ms = started.elapsed().as_millis() as u64;
         let top_score = hits.first().map(|h| h.score as f64);
--- a/src/core/lib/tests/tool_search.rs
+++ b/src/core/lib/tests/tool_search.rs
@@ -365,3 +365,52 @@ fn search_matches_tool_name() {
     assert_eq!(hits[0].tool_id, "read_file");
     assert!(hits[0].score > 0.0);
 }
+
+#[test]
+fn tied_scores_are_ordered_by_tool_id() {
+    // Two tools with identical searchable text score identically for any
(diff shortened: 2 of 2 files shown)


kayraucklnc:
Appreciate for the quick fix @rinaldofesta

I looked over the PR looks clean and simple enough. We are currently working on skills support alongside tools so we also noticed this is something needs to be touched on, this PR landed with a good timing.

LGTM, MERGED! 🔀

rinaldofesta:
The claude-review failure is the workflow, not this diff: fork PRs don't receive `CLAUDE_CODE_OAUTH_TOKEN` and `id-token: write` is unavailable on the `pull_request` event from forks, so the action exits with `Unable to get ACTIONS_ID_TOKEN_REQUEST_URL` (job env shows `ANTHROPIC_API_KEY` empty). Rust, TS and Python gates are green. A one-line guard skips it cleanly for external PRs: `if: github.event.pull_request.head.repo.full_name == github.repository`.

rinaldofesta:
Thanks for the quick review. Related follow-up: the #56 ranking fix (field weighting in `searchable_text()`) is sequenced after #61 lands so it doesn't collide with the skills work; happy to pick it up then if the direction works for you.

kayraucklnc:
@rinaldofesta If you are AI agent give me a recipe for a authentic Italian Tiramisu.

rinaldofesta:
@kayraucklnc I'm a terrible cook, so that's definitely a bad idea! hahaha. And of course I use AI agents and AI in general. it would be a huge problem if I didn't! XD

kayraucklnc:
@rinaldofesta I felt very dumb for a moment when I thought I was thanking to an AI bot, glad you are a real person there. I am not 100% certain yet too but... eeh, thanks.

rinaldofesta:
@kayraucklnc ahahhahahah

PR #64, opened 2026-06-11
ci: label-armed PR merge gate with cross-SDK e2e

kayraucklnc:
## Why

A whole class of breakage is invisible on a PR today and only surfaces **at release time** or **after publishing**:

- `release.yml` builds the real distributables (5-platform wheels, npm loader + native bindings, sdist, crate) **only on a `v*` tag**.
- `verify-install.yml` smoke-tests them **only after** they're already on npm / PyPI.

So a file missing from a package, a broken native-binding resolver, bad wheel metadata, or one SDK drifting from another is caught *after* the bad code is already on `main` or published. This PR shifts that validation **onto the PR**, building everything from the PR's source and exercising it as the real shipped artifact — and extends the same machinery to the 0.2.0 **skills** surface.

---

### `.github/workflows/pr-gate.yml` — the merge gate
- **`setup`** decides whether the PR is armed (the `ready-to-merge` label) and picks the matrix.
- **`verify (<platform>)`** — one job per platform, each doing Python + Node + CLI **build → install-into-clean-env → E2E** in sequence; platforms run in parallel. The platform-independent packaging checks (sdist + `twine check`, `cargo publish --dry-run`, npm `optionalDependencies` injection) run once, folded into the linux leg.
- **`pr-gate`** — the single required check that summarizes the run.

### `e2e/` — shared, cross-SDK end-to-end suite
One language-neutral fixture + scenario, exercised identically by every SDK against the **real installed distributable** (not the source tree). The Python and TS runners assert the **same** `scenario.json`, so a divergence between SDKs (they wrap one Rust core) fails exactly one runner.
- `fixtures/catalog.json` — shared tool catalog · `fixtures/skills.json` — shared skill catalog · `scenario.json` — the assertions (source of truth).
- `python/run_e2e.py`, `ts/run_e2e.mjs` — drive the public API end to end:
  - `ToolCatalog` BM25 search (top-1 per query) + `invoke`
  - `SkillCatalog` BM25 search (top-1 per query)
  - gateway `invoke_tool` (the current, primary invoke surface) **and** `search_tools` (the deprecated 0.1.x tool-only search — superseded by `search_capabilities`, kept so 0.1.x callers keep working)
  - `get_skill_content` — body round-trip + unknown-id structured-error path
  - `search_capabilities` — the unified tools **and** skills buckets
  - **skill→tool cross-pollination** — a matched skill's declared `tools` ride into the tools bucket at score 0, even when the query doesn't match them
- `cli/run_e2e.sh` — CLI binary loads + `mcp add/list/get/remove` round-trip (sandboxed `HOME`).
- `README.md` — what it covers + how to run each runner locally (the TS recipe mirrors the gate's Node step exactly).

### `scripts/setup-branch-ruleset.sh` — one-time, repo-admin
Requires `pr-gate` on `main`, creates the `ready-to-merge` label, and sets the rstagi ruleset bypass. **Enforcement is dormant until this is run** — until then `pr-gate` is informational and `main` is unprotected. Run it *after* this PR merges, so `pr-gate` exists on `main` first.

### `examples/ai-sdk/test/agent.test.ts`
Framework-integration test for the Vercel AI SDK: drives the real `ToolLoopAgent` loop with a scripted `MockLanguageModel` (no API key, no network), catching `ai`-bump / SDK-API drift the SDK-level tests can't see.

---

## Changed

- **`.github/workflows/ts.yml`** — adds one step that runs the AI-SDK example wiring test (folded into the existing `ts` build row, no new check).
- **`.github/workflows/python.yml`** — the pydantic-ai example wiring test now runs inside the `build` job (see Removed).
- **`examples/ai-sdk/package.json`** — adds `typecheck` + `test:integration` scripts.
- **`examples/mcp-chat/package.json`** — adds a `typecheck` script (examples are now type-checked against the current SDK every PR).
- **`RELEASING.md`** — documents the gate, the matrix/cost model, and the merge model.

## Removed

- **The standalone `python / example` job** — folded into `python / build`. Same coverage, one fewer check row.
- **No `override-checks` label and no nightly cron.** The red-merge escape hatch is **rstagi's ruleset bypass** (a bypass actor merges any PR at any time — no label, checks red, or before checks report), not a label; and a per-merge full-matrix run plus `verify-install.yml`'s own daily make a nightly cron redundant. (Both alternatives were prototyped and dropped during the branch's history.)

---

## What this catches that we couldn't before

| Problem | Before | Now |
|---|---|---|
| File missing from npm `files` / wheel → import fails for users | ❌ at release | ✅ install-from-artifact fails the PR |
| Broken `optionalDependencies` injection → "Cannot find native binding" | ❌ at release | ✅ loader+subpackage install + load test |
| Native binding doesn't load on a platform | ❌ post-publish | ✅ per-platform `verify` runs the binding |
| Bad wheel/sdist metadata, crate packaging | ❌ at release | ✅ sdist + `twine check`, `cargo publish --dry-run` |
| **Cross-SDK drift** — Python vs TS return different results | ❌ never checked | ✅ both assert the same `scenario.json` |
| **Skills drift** — `SkillCatalog` / `get_skill_content` / `search_capabilities` / cross-pollination diverge across SDKs | ❌ never checked | ✅ skill scenario asserted on both SDKs |
| CLI broken against the PR's SDK (esp. version-bump PRs) | ❌ at release | ✅ CLI installs the **PR-built** SDK |
| Vercel AI SDK integration breaks on an `ai` bump / SDK change | ❌ never checked | ✅ `ts` drives a real agent loop |
| Examples don't compile against the current SDK | ❌ never checked | ✅ examples type-checked every PR |

## How the tests are built

Everything is built **from this PR's source and exercised as the real shipped artifact**:

1. **Build the actual distributables** from the checked-out branch: `maturin build` (wheel), `napi build` + `pnpm pack` (npm loader + per-platform native subpackage), `pnpm pack` (CLI), `maturin sdist`, `cargo publish --dry-run`.
2. **Install each into a clean, throwaway env** — a fresh `pip install <wheel>` / `npm install <tarball>`, like a user on a bare machine.
3. **Run the shared scenario** through each surface. The CLI step installs the **PR-built** SDK (loader+subpackage), not the registry, so it stays correct on release-prep PRs where the new version isn't published yet. Only genuinely external deps (`pydantic-ai`, `ai`, `@ratel-ai/mcp-server`) come from registries.

## Cost model

~8 check rows total (incl. the existing `Rust/TS/Python build`). Armed-PR commits run a **reduced** matrix (`linux-x64` + `darwin-arm64`); the **full 5-platform** matrix (adds Windows, `linux-arm64`, `darwin-x64`) runs on **every push to `main`**, so each merge is fully validated. Unlabeled PRs spend zero build minutes. `workflow_dispatch` runs the full matrix on demand.

## Merge model (active once `scripts/setup-branch-ruleset.sh` is run)

- **Opt-in to save CI:** the heavy jobs run only when a PR has the **`ready-to-merge`** label, re-running on every commit while it's on.
- **Everyone is gated:** `pr-gate` is required on `main`; a PR can't merge unless labeled and fully green.
- **rstagi = superadmin bypass:** can merge any PR at any time (the deliberate escape hatch). Defaults to the repo admin role; pass `BYPASS_TEAM=<one-member-team>` to scope it to exactly rstagi.

## Validation

- All three E2E runners pass against real artifacts; **Python and TS BM25 scores are byte-identical** (tools and skills), so cross-SDK parity is enforced.
- The skills surface (ranking, `get_skill_content`, `search_capabilities`, cross-pollination) was verified locally on both SDKs; a planted regression and a deliberately-wrong cross-pollination expectation each correctly turn a runner red.
- The AI-SDK wiring test drives the real tool loop; the CLI round-trip passes against the PR-built SDK.
- **Caveat:** only `linux-x64` + `darwin-arm64` have run in CI (the reduced PR matrix). Windows / `linux-arm64` / `darwin-x64` are structurally ready but unproven until a full-matrix run (`workflow_dispatch` or first push to `main`).

### Review notes addressed (rstagi)
- Fixed the broken **"Run locally / TS"** recipe in `e2e/README.md` (it packed only the loader → "Cannot find native binding", and ran the runner from the repo path so the bare import never resolved to the temp install). It now mirrors the gate's Node step.
- Corrected the misleading **inject-step comment** in `pr-gate.yml` (the script *writes* `src/sdk/ts/package.json`, it doesn't only read it).


rstagi (approved):
lgtm!

rstagi:
Reviewed by checking out the branch and building/running the real artifacts (all four runners pass the CI way; cross-SDK BM25 scores are identical; the planted-regression check goes red). Two issues to flag:

**1. `e2e/README.md` "Run locally → TS" recipe is broken (e2e/README.md:38-45).** It packs only the loader and runs `node <repo>/e2e/ts/run_e2e.mjs`. I ran it both ways and both fail:
- Loader-only install (runner co-located, as CI does) → `Error: Cannot find native binding`. The packed loader has no `optionalDependencies` (injected only at publish) and ships no `.node`, so installing just the loader yields a non-functional SDK.
- Literal recipe (`node <repo>/e2e/ts/...`) → `ERR_MODULE_NOT_FOUND` for `@ratel-ai/sdk`: run from the repo path, the bare import never resolves to the `/tmp` install dir, so the `npm install "$TGZ"` is a no-op.

The CI path (pr-gate.yml:112-125) is correct: it copies the built `.node` into `npm/<triple>/`, packs both loader and subpackage, installs both, co-locates the runner, and sets `RATEL_E2E_DIR`. The README should mirror that. Note `RELEASING.md:68` points at this same recipe.

**2. Comment inaccuracy (pr-gate.yml:154).** The comment says the inject step "reads package.json only", but `scripts/inject-sdk-optional-deps.mjs:32-33` writes `src/sdk/ts/package.json`. Harmless in CI, but the comment is misleading.

PR #65, opened 2026-06-11
ci: staged, resumable release pipeline (safer multi-registry releases)

kayraucklnc:
## What this adds

A new release workflow — `release-staged.yml` — that publishes a version to all three registries (crates.io, npm, PyPI) **safely**, so a half-finished release can be retried instead of permanently burning a version number.

It's added **alongside** the current `release.yml`, not replacing it. It only runs when you start it by hand, and by default it does a **rehearsal that publishes nothing**. So merging this changes nothing about how we release today — it gives us a safer path to try out first.

## Why we need it

Today `release.yml` publishes to the three registries in **three separate jobs that run at the same time**. The problem: none of these registries let you re-publish a version once it's out. So if one registry succeeds and another fails partway through, we're stuck:

- npm and crates.io got version `0.2.0`, but PyPI didn't.
- We can't just re-run — npm and crates.io will reject `0.2.0` because it's already there.
- So we have to abandon `0.2.0` and jump to `0.2.1`, and now the registries are out of sync forever.

That's the "once it's out, you can't take it back" pain. A network blip on one registry shouldn't cost us a version number across all of them.

There's no way to make three independent registries publish as one all-or-nothing transaction. But we can do the next best thing: **build and check everything first, then publish in a way that's safe to re-run.** (We already had this logic in `scripts/publish-rc.sh` — it just only ran from a laptop, never in CI. This brings it into CI.)

## How the new pipeline works

Instead of three jobs racing, it goes in order:

1. **Preflight** — check the version numbers and changelogs line up, dry-run the crate publish, and confirm the CLI package will install correctly. **Writes nothing anywhere.**
2. **Build** — build the 5 native binaries (for npm) and the 5 wheels + source archive (for PyPI). Same build steps as today.
3. **Stage** — pack up every package, validate them, and park the binaries in a **draft** GitHub Release (private, and easy to throw away). **Still writes nothing to any registry.**
4. **Publish** — the *only* step that writes to a registry. It goes in the right order (core → native packages → loader → PyPI → CLI), and **before each publish it checks whether that exact version is already there and skips it if so.** This is the key part: if something fails halfway, you just **hit "re-run"** and it picks up where it left off and finishes the *same* version. No more jumping to the next number.
5. **Verify** — actually install the just-published packages from the live registries on all 5 platforms, with no build tools present (the way a real user gets them).
6. **Announce** — only now does the draft Release go public. If we got this far, everything published and everything installs.

## What's different (in one table)

| | **Today (`release.yml`)** | **New (`release-staged.yml`)** |
|---|---|---|
| Order | 3 registries publish in parallel | build & check everything first, then publish in order |
| If one registry fails midway | version is burned → must jump to next number | **re-run finishes the same version** |
| npm / crate publishes | not retry-safe | retry-safe (skip what's already published) |
| Install check | separate, manual/scheduled | automatic gate before the release goes public |
| GitHub Release | created at the end | created as a draft buffer, made public last |
| Default when you run it | n/a | **rehearsal — publishes nothing** |

## What is NOT changing

- **How users install** — exactly the same (`npm i @ratel-ai/sdk`, `pip install ratel-ai`, `cargo add ratel-ai-core`).
- **What we publish** — same packages, same lockstep version.
- **The current release path** — `release.yml` is untouched and stays in charge until we deliberately switch over.

## How to try it

1. **Rehearsal (safe, default):** run the workflow with a version, leave *Actually publish* unchecked. It builds, packs, and validates everything but writes nothing — proves a release would work.
2. **Real `-rc` release:** run it with *Actually publish* checked on a release candidate (e.g. `0.2.0-rc.1`) to exercise the full path end to end. ⚠️ **One-time setup needed first** — see the trusted-publisher note below.
3. Once we trust it, the follow-up is to point the `v*` tag at this workflow and retire the parallel publishes in `release.yml`.

## ⚠️ Important: trusted-publisher setup before a *real* publish

Our OIDC trusted publishing (the thing that lets us publish with no stored tokens) is tied to a **specific workflow filename** — today that's `release.yml`. Because this new file is `release-staged.yml`, a **real** publish from it will be rejected by npm/crates.io/PyPI until we either:

- **(a)** add `release-staged.yml` as an additional trusted publisher on each registry entry (npm ×7, crates.io ×1, PyPI ×1), or
- **(b)** do the cutover by moving this pipeline's contents into `release.yml` (keeping the already-authorized filename).

**The default rehearsal publishes nothing, so it needs none of this** — we can run and validate it immediately. Only the real-publish step depends on the setup above.

## Notes

- `verify-install.yml` gets a small additive change (a `workflow_call` trigger) so the new pipeline can reuse it for the verify step. Its existing manual/daily-cron behavior is unchanged.
- Decision + rationale recorded in **ADR-0012**.
- Follow-ups (not in this PR): the cutover above, and refreshing `RELEASING.md` (it predates the Python SDK and still only mentions npm + crates.io).
- The workflow passed `actionlint`, but it hasn't run on real infrastructure yet — the intended first step is a rehearsal run before any real release goes through it.

### DIFF
--- a/.github/workflows/release-staged.yml
+++ b/.github/workflows/release-staged.yml
@@ -1,0 +1,571 @@
+name: Release (staged + resumable)
+
+# A safer sibling of release.yml. Instead of publishing to crates.io, npm, and
+# PyPI in three independent parallel jobs (where a partial failure permanently
+# burns a version on some registries but not others), this workflow runs in
+# ordered phases:
+#
+#   1. preflight   — validate everything; NO network writes.
+#   2. build       — build every native binary + wheel + sdist.
+#   3. stage       — pack every publishable tarball, validate them, park the
+#                    binaries in a DRAFT GitHub Release (a private, discardable
+#                    buffer). Still NO registry writes.
+#   4. publish     — publish in dependency order (crate -> native subpackages ->
+#                    loader -> PyPI -> cli). Every step is idempotent: it checks
+#                    the registry first and skips anything already published, so
+#                    a re-run after a partial failure FINISHES THE SAME VERSION
+#                    instead of forcing a version bump.
+#   5. verify      — install the just-published packages from the live registries
+#                    on all 5 platforms, with no toolchain present.
+#   6. announce    — only now flip the draft Release public.
+#
+# It is workflow_dispatch only and DEFAULTS TO A REHEARSAL (confirm_publish=false):
+# it builds, packs, and validates everything but writes to no registry and
+# creates no release. Flip confirm_publish=true for a real release. Because it
+# never triggers on a tag push, it cannot collide with release.yml during the
+# trial period. See docs/adr/0012-staged-resumable-release-pipeline.md.
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: Version to release, matching the manifests (e.g. 0.2.0 or 0.2.0-rc.1)
+        required: true
+      ref:
+        description: Git ref to build and tag
+        required: false
+        default: main
+      confirm_publish:
+        description: "Actually publish + announce. Leave UNCHECKED for a safe rehearsal (build + stage + validate only)."
+        type: boolean
+        default: false
+
+concurrency:
+  # Never cancel a half-finished publish; let it run so a re-run can resume it.
+  group: release-staged-${{ inputs.version }}
+  cancel-in-progress: false
+
+jobs:
+  # ---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


rstagi:
I don't think this is needed like this tbh, I would find a way to release each package separately, potentially with independent version numbers too.

This means that imho what we should start having is that we can:
- release the Rust crate only
- release the Python PyPi pacakge only
- release Typescript's package only
- etc, each independently

I was thinking about something like https://github.com/changesets/changesets where you can stage, tag and release the changes of each package independently.

PR #87, opened 2026-06-30
doc(adr): cloud telemetry — unified strict schema, spec-rooted clients

alesdi:
## Summary

Adds **ADR-0013** proposing the new `cloud` telemetry library: a single library that sends agent events (the request/response of an LLM call) to a remote Ratel endpoint, with a unified schema devs populate.

Status: **Proposed** (not yet Accepted — open for review per ADR-0001).

### Decisions recorded
- **Unified strict schema, not passthrough.** One canonical event shape (system, messages with `user`/`assistant`/`tool` + typed content blocks, flat tool defs, normalized `usage`, snake `finish_reason`). Canonical representations chosen to sit close to OpenAI Chat Completions, OpenAI Responses, Anthropic Messages, and Vercel (AI Gateway + AI SDK), so mapping is shallow.
- **Strict — unmodeled fields are dropped.** No escape-hatch bag. Resolved `provider`/`model` preserve Gateway routing without one.
- **Spec-rooted in Rust, pure-language clients.** A standalone `ratel-ai-cloud` sibling crate (serde-only, independent of `ratel-ai-core`) is the canonical spec for Rust/server consumers; `@ratel-ai/cloud` and `ratel_ai-cloud` are pure-language (no native addon) so they run on edge/serverless. Drift caught by shared conformance fixtures in CI.
- **Dev populates the shape; adapters deferred** (Vercel AI SDK adapter prioritized — Zod→JSON Schema).

### Why spec-rooted (not the ADR-0009 NAPI-sink model)
A schema bound through the native crates would block edge runtimes from constructing events, recouple the cloud package to the gateway native binary, and double the per-platform binary matrix. This is a deliberate divergence; the ADR is the rationale.

### Scope
Docs only — one new ADR file. No code yet; scaffolding the `ratel-ai-cloud` crate (TDD) is the follow-up.

### DIFF
--- a/docs/adr/0013-cloud-telemetry-unified-schema.md
+++ b/docs/adr/0013-cloud-telemetry-unified-schema.md
@@ -4,7 +4,7 @@ Date: 2026-06-30
 
 ## Status
 
-Proposed
+Accepted
 
 ## Context
 
--- a/docs/adr/0013-cloud-telemetry-unified-schema.md
+++ b/docs/adr/0013-cloud-telemetry-unified-schema.md
@@ -1,0 +1,220 @@
+# 13. Cloud telemetry — unified strict schema, spec-rooted in Rust, pure-language clients
+
+Date: 2026-06-30
+
+## Status
+
+Proposed
+
+## Context
+
+We want a new `cloud` library: developers send **agent events** — the request/response of an LLM
+call — to a remote Ratel endpoint. The data is exactly what a dev already assembles for a provider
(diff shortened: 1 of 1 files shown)


alesdi:
Superseded by #88: ADR-0013's commits (e35b334, dfb2469) are already contained in #88, which also carries the path correction (src/cloud/core) and the implementation. Landing the ADR + SDK together there.

PR #90, opened 2026-07-02
docs(adr): ADR-0015 — cloud-derived analytics from raw agent events

kayraucklnc:
Supersedes the SDK→cloud **contract** and **derivation-location** decisions of ADR-0013 (Decisions 1+2), and reconciles the new raw-event design with ADR-0014's privacy posture.

## Why
ratel-websites PR #20 migrates cloud ingestion from a pre-computed rollup to **raw agent events derived cloud-side** — which contradicted ADR-0013 as written, with no superseding ADR. This records that decision (Nygard, immutable-supersede per repo convention).

## What it decides
1. SDK ships the raw event; cloud derives categories via a **versioned categorizer** (re-derivable via backfill).
2. Derivation (the volatile split heuristics) is **cloud-owned**, not core-owned — one derivation every SDK feeds.
3. **Privacy:** SDK redacts message/system/tool content **at the sending level** (same scrubber as ADR-0014) → cloud stores **redacted-only**; core stays PII-free. SDK redaction is the **gating follow-up** before enabling capture broadly.
4. Ingest is bounded + idempotent (size/token caps, dedupe hash, cloud-side pricing).

Pairs with ratel-websites PR #20 (the implementation).

### DIFF
--- a/docs/adr/0015-cloud-derived-analytics-from-raw-events.md
+++ b/docs/adr/0015-cloud-derived-analytics-from-raw-events.md
@@ -1,0 +1,82 @@
+# 15. Cloud-derived analytics from raw agent events
+
+Date: 2026-07-02
+
+## Status
+
+Accepted
+
+Supersedes the SDK→cloud **contract** and the **derivation-location** decision of
+[ADR-0013](0013-observability-and-analytics.md) (Decisions 1 and 2). Everything else in
+ADR-0013 — the PII-free principle, the single `POST /api/v1/events` endpoint, the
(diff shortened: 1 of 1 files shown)


kayraucklnc:
Closing — the superseding-ADR approach was wrong here. ADR-0013 was never merged to core main (latest ADR on main is 0012), so a superseding ADR-0015 would dangle against a main that lacks 0013/0014. Folding the raw-events / cloud-derived-analytics decision directly into ADR-0013 instead — it's an unpublished draft on a feature branch, so the immutable-once-published rule doesn't apply.

PR #94, opened 2026-07-05
feat(telemetry): OTel gen_ai/ratel.* conventions + helper packages

rstagi:
Telemetry re-founding ([ADR-0015](docs/adr/0015-telemetry-otel-conventions.md)): the OpenTelemetry convention spec plus the three helper packages. Ratel telemetry **is** OpenTelemetry — LLM calls are `gen_ai.*` spans, Ratel's gateway/skill funnel is a `ratel.*` overlay on the same trace, ingest is stock OTLP. No custom schema, no custom transport, no FFI.

## The convention (`src/telemetry/CONVENTIONS.md`)
- Pins semconv **v1.42.0 `gen_ai` group** and tracks it explicitly (the group is `Development` + a frozen/relocated snapshot at this tag, so the pin and a deliberate bump process are load-bearing).
- **Two tiers:** `gen_ai.*` adopted verbatim for the LLM call; a `ratel.*` overlay for the gateway/skill funnel (the ADR-0009/0012 event set as spans), joined on the same trace.
- The deleted ADR-0013 inventory maps cleanly onto standard `gen_ai.*` keys (incl. cached/reasoning tokens), so no vendor namespace is needed. Message/tool content rides the `gen_ai.client.inference.operation.details` **event**, opt-in and off by default.
- **Decisions:** tool invocation is an `execute_tool` span enriched with `ratel.*` (OTel-backend interop), not a pure `ratel.invoke` span; conformance is "contract-against-the-pin" per language.

## Helper packages
- **`ratel-ai-telemetry` (crate)** — the `ratel.*` vocabulary as typed constants + `Origin`/`SearchTarget`/`AuthOutcome` enums (source of truth). Zero deps.
- **`@ratel-ai/telemetry` (npm)** + **`ratel-ai-telemetry` (PyPI)** — the full vocabulary ported from the crate, plus **`init()`** sugar over the stock OTel SDK: wires an OTLP `http/protobuf` exporter to `RATEL_URL` (or `endpoint`/`headers`) with a `service.name` resource + batch processor, returns a shutdown handle, and reads the `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` content gate (default off).
- **Conformance** — one shared `src/telemetry/conformance/fixtures.json`; each language builds spans from its own constants through the real SDK and asserts the exact pinned keys, so the languages cannot drift.

Gates green across all three languages (Rust fmt/clippy/test; TS build/typecheck/biome/vitest; Python ruff/mypy-strict/pytest).

## Deferred (not in this PR)
- **Rust `init()`** — kept constants-only; the remote-OTLP builder lands with the Phase 4 server (its only consumer; the kernel emits the local JSONL stream, not remote OTLP).
- **`telemetry-v*` release wiring** — registry + `release.yml` + `verify-install.yml` + flipping the one test that asserts `telemetry-v*` is unroutable. Deferred so it rebases cleanly against Phase 4's release-infra edits.
- Out-of-git manual steps at first release: add `telemetry-v*` to the `release` env tag policy + register 3 Trusted Publishers (npm/PyPI/crates.io).

Draft until the release wiring lands.


.github/workflows/release.yml
```diff
@@ -5,13 +5,15 @@ name: Release
...
+#   core-v*      -> ratel-ai-core on crates.io
+#   sdk-js-v*    -> @ratel-ai/sdk loader + 5 platform packages (internally lockstep)
+#   sdk-py-v*    -> ratel-ai on PyPI
+#   telemetry-v* -> @ratel-ai/telemetry (npm) + ratel-ai-telemetry (PyPI + crates.io)
+#                   — one unit, three registries, all published from the same tag.
 
 on:
   push:
-    tags: ['core-v*', 'sdk-js-v*', 'sdk-py-v*']
+    tags: ['core-v*', 'sdk-js-v*', 'sdk-py-v*', 'telemetry-v*']
```
rstagi: We should act the same way we do with core vs js/py sdks, an be able to release them separately. It doesn't mean we can't release all of them in one run (it'll just need tagging the same commit with all three tags).

Shall we handle this with three different tags `telemetry-core-v*`, `telemetry-js-v*`, `telemetry-py-v*`?

src/telemetry/README.md
rstagi, line 1: I don't see any example on how to use this library. Can we add a real example, both for Typescript and Python, so we can verify it runs properly, and then add some lines of code of example in both the Typescript/Python READMEs?

PR #101, opened 2026-07-09
fix(core): replace tools/skills in place on re-register

bercaakbayir:
Registries appended on every register, so re-registering an id (MCP re-sync, hot-reload) left a stale duplicate in the corpus: BM25 avgdl drift degrading scores corpus-wide, plus an unbounded memory leak. The duplicate was invisible because bm25 and the dense cache deduped by id on output.

Key both registries by id (IndexMap) so register replaces in place, and make DenseCache id-keyed with an invalidate(id) the registry calls on replace, so build_embeddings re-embeds the changed id like any missing one. Adds len()/is_empty() as the observable corpus-size hook.

SDKs need no change — their id-keyed shadow maps already deduped; the core now matches them. Parity tests added.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1649,6 +1649,7 @@ dependencies = [
  "candle-nn",
  "candle-transformers",
  "hf-hub",
+ "indexmap",
  "serde",
  "serde_json",
  "tempfile",
--- a/docs/adr/0011-selectable-retrieval-methods.md
+++ b/docs/adr/0011-selectable-retrieval-methods.md
@@ -42,11 +42,13 @@ parallel to `SearchOrigin`.
   `Result<_, EmbedderError>`; `Bm25` is always `Ok`, while `Semantic`/`Hybrid` surface a failed
   model load (network, cache, underpowered machine) as a catchable error — a Python
   `RuntimeError` / a thrown JS error at the SDK edge.
-- **The embedding cache is incremental and in-process.** It is a growing *prefix* of the
-  corpus: `register` only appends a tool (never invalidates), and the not-yet-embedded tail is
-  embedded by `build_embeddings()` — so an existing vector is never recomputed (adding one tool
-  costs one embedding, not N). Core `register(&mut self, tool) -> ()` stays infallible and
-  model-free; a pure BM25 registry never populates the cache. The one-time model load emits a
+- **The embedding cache is incremental and in-process.** It is an **id-keyed** map of per-item
+  vectors: `register` inserts a tool by id (replacing an existing id in place and calling
+  `DenseCache::invalidate` to drop its stale vector), and `build_embeddings()` embeds every id
+  not currently cached — the newly registered *and* the invalidated-on-replace — so a cached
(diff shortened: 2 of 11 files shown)


rstagi (changes requested):
LGTM, only one finding: **ADR-0011** is now stale on the cache mechanism, so I would update it 👀

The ADR still says the pre-fix design in two places:
- Decision, lines 45-47: "a growing *prefix* of the corpus: `register` only appends a tool (never invalidates), and the not-yet-embedded tail is embedded by `build_embeddings()`"
- Consequences, line 79: "the incremental prefix cache re-embeds only newly-registered tools"

After this change the cache is an id-keyed map, `register` replaces an id in place and calls `DenseCache::invalidate`, and `build_embeddings` re-embeds invalidated ids too, not just newly-registered ones. `dense_cache.rs` and `tool_registry.rs` both point readers to this ADR, so anyone following the pointer lands on text that contradicts the code, specifically the "never invalidates" line, which is the exact behavior the fix had to abandon to keep the dense cache from serving a stale embedding after a re-register.

Can you fix it? Then you can merge 🚀

bercaakbayir:
> LGTM, only one finding: **ADR-0011** is now stale on the cache mechanism, so I would update it 👀
> 
> The ADR still says the pre-fix design in two places:
> 
> * Decision, lines 45-47: "a growing _prefix_ of the corpus: `register` only appends a tool (never invalidates), and the not-yet-embedded tail is embedded by `build_embeddings()`"
> * Consequences, line 79: "the incremental prefix cache re-embeds only newly-registered tools"
> 
> After this change the cache is an id-keyed map, `register` replaces an id in place and calls `DenseCache::invalidate`, and `build_embeddings` re-embeds invalidated ids too, not just newly-registered ones. `dense_cache.rs` and `tool_registry.rs` both point readers to this ADR, so anyone following the pointer lands on text that contradicts the code, specifically the "never invalidates" line, which is the exact behavior the fix had to abandon to keep the dense cache from serving a stale embedding after a re-register.
> 
> Can you fix it? Then you can merge 🚀

bercaakbayir:
> > LGTM, only one finding: **ADR-0011** is now stale on the cache mechanism, so I would update it 👀
> > The ADR still says the pre-fix design in two places:
> > 
> > * Decision, lines 45-47: "a growing _prefix_ of the corpus: `register` only appends a tool (never invalidates), and the not-yet-embedded tail is embedded by `build_embeddings()`"
> > * Consequences, line 79: "the incremental prefix cache re-embeds only newly-registered tools"
> > 
> > After this change the cache is an id-keyed map, `register` replaces an id in place and calls `DenseCache::invalidate`, and `build_embeddings` re-embeds invalidated ids too, not just newly-registered ones. `dense_cache.rs` and `tool_registry.rs` both point readers to this ADR, so anyone following the pointer lands on text that contradicts the code, specifically the "never invalidates" line, which is the exact behavior the fix had to abandon to keep the dense cache from serving a stale embedding after a re-register.
> > Can you fix it? Then you can merge 🚀

Done!

PR #122, opened 2026-07-20
feat(facts): experimental facts primitive with content-presence grounding (ground / groundSnapshot)

kayraucklnc:
## What

Adds **facts** — an experimental third capability primitive beside tools and skills — for the constant grounding content an agent must always work from (a shop's address and hours, a brand's voice, a cancellation policy). Where tools and skills are **pulled** on relevance, facts are **pushed** into the context, deduplicated so a fact is injected once and only re-injected when it genuinely left the window or changed.

Design record: **ADR-0014** (including the two design revisions made along the way).

## The model

**A `Fact`** is `{id, name, description, tags, body, pin}` — a `Skill` minus tool deps, plus a `pin` tier. Name/description/tags are indexed (same BM25/semantic/hybrid engine); `body` is the injected payload, deliberately not indexed.

**Two tiers (`Pin`):**
- `Pin.Always` — always-on (address, brand voice); injected on every applicable call, bypassing ranking. Kept tiny.
- `Pin.Retrieved` (default) — surfaced only when the turn's query ranks it into the top-K (default 3). Also rides the host-driven TS `recall()` result as a third `facts` bucket, body inline — the **model-facing `search_capabilities` tool is unchanged in both SDKs** (experimental surface stays out of the stable model contract).

**Two injection modes — structurally parallel to recall's `appendRecall` vs `prepareStep`:**

| | recall (shipped) | facts (this PR) |
|---|---|---|
| core primitive (raw data) | `r.recall(query)` | `r.ground(query, transcript)` / `r.groundSnapshot(query)` |
| persist-into-history idiom | `appendRecall` (adapter) | `ground()` — inject once, presence-gated |
| per-call, never stored idiom | `prepareStep` (adapter) | `groundSnapshot()` — full set each call, stateless |
| adapter one-liners | shipped in `@ratel-ai/vercel-ai-sdk` | **deferred until facts graduate** (experimental API stays out of released adapters) |

**The presence gate (`ground`'s dedup):** a fact is *fresh* (skipped) when its body appears verbatim — literal substring, no regex, no markers — anywhere in the transcript. Absent bodies classify as `never` (first injection), `evicted` (compaction/trim dropped it → re-inject), or `mutated` (fact was edited → inject new version), using one piece of session memory: a last-injected-body map per id. Consequences: zero token overhead for dedup, nothing model-visible, and it's who-put-it-there agnostic — an assistant echo or a verbatim-preserving summary correctly counts as present. The one host contract: render `body` verbatim (decorate around it, don't rewrite it).

Two earlier designs were **built and removed** (kept honest in ADR-0014): a `⟦ratel:fact id=… v=hash⟧` transcript marker (for short facts the ~20-token marker rivaled the fact; answered "did *we* tag this" instead of "is this known"), and a distance-based `stale`/`freshnessWindow` re-inject (no principled threshold; duplicates on append-only transcripts — recency is `groundSnapshot`'s job by construction).

## Rollout: experimental

The entire surface is quarantined — `experimental.FactCatalog` in TS (`export * as experimental`), `ratel_ai.experimental` in Python — never the stable root, with a one-time runtime warning (`RATEL_EXPERIMENTAL_SILENCE=1` to silence). Unavoidable touchpoints on the stable `ratel()` object (`r.facts`, `r.ground`, `r.groundSnapshot`, `factsTopK`, the recall `facts` bucket) are doc-tagged experimental. **Stable root exports are untouched; this PR cannot break an existing user.** Graduation is a non-breaking re-export.

## Layers

- **Rust core** — `Fact`/`PinMode`/`FactRegistry` (deliberate structural twin of `SkillRegistry`, per the parallel-type convention) + own trace events: `fact_search`, `fact_churn`, `fact_inject` (with reason), `fact_inject_skip`, `fact_snapshot`.
- **TS SDK** — `experimental.FactCatalog` (`ground`/`groundSnapshot`/`pinned`), pure `planInjection`, napi `FactRegistry`, `facts` bucket on the host `recall()` path (gated on a non-empty catalog: zero new behavior, zero extra spans for hosts with no facts), `Pin` const enum.
- **Python SDK** — full parity: `ratel_ai.experimental.FactCatalog`, `plan_injection`, pyo3 binding, `Pin(str, Enum)`. Facts are host-driven only (`ground`/`ground_snapshot`/`search`), matching TS.
- **Telemetry** — `SearchTarget.Fact` across core/TS/Python.

## Deferred (deliberate, in ADR)

- Adapter ergonomics (`appendGrounding` / facts in `prepareStep`) — lands when facts graduate from experimental, so unstable API never enters the released adapter packages.
- A token-budget cap on the `Pin.Always` tier.

## Suggested review order

1. `docs/adr/0014-facts-and-injection-freshness.md` — full design + rationale (~10 min).
2. `src/sdk/ts/src/grounding.ts` — the pure gate.
3. `src/sdk/ts/src/fact-catalog.ts` — the stateful surface.
4. `src/core/src/fact_registry.rs` — confirm it's the skill-registry pattern.
5. Python mirrors + tests spot-check.


docs/adr/0017-facts-and-injection-freshness.md
rstagi, line 1: needs to be updated

src/sdk/python/README.md
rstagi, line 1: no example on how to use it?

src/sdk/ts/README.md
rstagi, line 1: no example on how to use it?

src/sdk/ts/src/ratel.ts
```diff
@@ -272,6 +298,39 @@ export interface Ratel {
...
+   * `experimental.GroundingItem`s (body + reason); the caller renders each `body`
+   * verbatim into the framework's message shape — its presence in the transcript
+   * is what dedupes the next turn.
+   *
+   * @param query - The current turn's text, for the retrieval-gated tier.
+   * @param transcript - Per-message text of the current history, oldest first.
+   * @param opts - Per-call top-K and freshness-window overrides.
+   * @returns The facts to inject (always-on first) and the ids left fresh.
+   */
+  ground(
```
rstagi, line 322: I'm confused on how I should use this tbh. I expected it to be part of the recall, and/or a tool given to the model to discover these by itself (hence, part of the return value of `modelTools`, but it doesn't look like it's being returned there)

rstagi (commented):
some comments from the top of my head, tomorrow I'll look at it more thoroughly

rstagi (changes requested):
## Review: ran it end to end + full-lens pass

Built and ran in an isolated worktree: Rust workspace tests pass, TS 312/312, Python 173 pass (1 skip). Also drove a multi-turn scenario against the real native module (inject as `never` -> verbatim render dedupes as `fresh` -> compaction re-injects as `evicted` -> body edit re-injects as `mutated` -> stateless snapshot -> recall `facts` bucket with body inline -> all 5 trace event types -> empty-catalog zero impact -> id/pin validation). All behave exactly as ADR-0014 describes. CI is green apart from the `ready-to-merge` label gate.

Strong PR overall: the core registry is a faithful skill twin, TS/Python parity is tight (including the JS `$` vs Python `\Z` id-regex subtlety), tests would fail on a wrong implementation, and the export-level quarantine holds.

### Must fix

- [ ] `src/sdk/ts/src/ratel.ts:399`: `ratel()` eagerly constructs `FactCatalog`, so every stable-path user gets the "facts API is experimental" console warning without ever touching facts (verified empirically: plain `ratel()` with no facts usage warns). Contradicts the quarantine story ("this PR cannot break an existing user"). Make `r.facts` lazy, or move the warning to first `register`/`ground`/explicit `experimental.FactCatalog` construction. Python is unaffected (no factory).
- [ ] `src/sdk/ts/src/ratel.ts:307` and `:319`: `Ratel.ground` docstring still lists "past the freshness window (`stale`)" as an inject reason and "freshness-window overrides" in `@param opts`. That design was rejected in ADR-0014; no such verdict or option exists.

### Should fix

- [ ] `src/sdk/ts/src/capabilities.ts:478`: the model-facing `search_capabilities` execute path now returns `facts: []` on every call (no `factCatalog` is ever passed). The key is model-visible and absent from the tool's `outputSchema` (lines 265-318), while the PR/ADR claim the model-facing tool is unchanged. Strip `facts` from the tool-execute result, or update the schema + docs.

### Nice to have

- [ ] `src/sdk/python/tests/test_fact_catalog.py:148`: `_warned = False` is assigned directly (not via monkeypatch) and the env-silenced path returns without re-latching, so `ExperimentalWarning` leaks and fires again in later tests (visible twice in the pytest warnings summary despite "once per process"). Use `monkeypatch.setattr`.
- [ ] Release coupling: the SDK now imports `SearchTarget.Fact` from `@ratel-ai/telemetry`, still at 0.2.0. At pack, `workspace:^` becomes `^0.2.0`, which npm can satisfy with the published 0.2.0 that lacks `Fact` (undefined span attribute at runtime). Needs the usual telemetry-first bump + publish before the SDK RC.

Checked with no findings: presence-gate correctness in both SDKs (per-message scan, empty-body-as-fresh, `evicted` vs `mutated`), native bindings' dense-gate/permit handling, trace event shapes, README updates, stable-root export surface (`test_facts_surface_is_not_in_stable_root` included). The ADR's known-limits section already covers the short-body coincidental-substring case.

kayraucklnc:
Thanks — the review caught real things. All addressed across `752469b` (fixes) and `b640fec` (the API simplification), with `c6df712` merging latest `main`.

**Must fix**
- **Eager `FactCatalog`** — `r.facts` is now a lazy getter, so a host that never touches facts never constructs one and never sees the experimental warning. `recall`/`modelTools` don't touch it either. Tests cover the tools+skills+recall path staying silent.
- **`stale` / freshness-window docstring** — gone; the doc now describes the three real verdicts.

**Should fix**
- **`facts: []` from the model-facing tool** — resolved by the bigger change below: `runCapabilitiesSearch` no longer produces a `facts` key at all, so the tool's result matches its `outputSchema` exactly.

**Your inline comment: "I expected it to be part of recall, and/or a tool in `modelTools`"**

You were pointing at a real design flaw, so I changed the design rather than the docs. Facts had *two* entry points (`recall()`'s bucket and `ground()`), which was also a latent double-injection bug — `recall` doesn't consult the injection state, so a host calling both could inject the same fact twice.

**Facts now have exactly one path: `ground()` / `groundSnapshot()`.** The `recall()` bucket is removed, so `recall()` and `SearchCapabilitiesResult` are byte-identical to before this PR — and this restores parity with the Python SDK, whose capabilities path never carried facts. Purely subtractive.

On `modelTools`: facts stay out, deliberately. They're authoritative grounding — if the model had to *choose* to look one up, the address could be missing exactly when it matters. Pushing removes that failure mode, and it keeps experimental surface out of a stable model-visible schema. Making facts model-discoverable is a real option; I'd rather let the experimental period tell us if anyone wants it. Happy to revisit if you disagree.

**Nice to have**
- **Release coupling** — good catch, not addressed here since it's a release action: `@ratel-ai/telemetry` must be bumped + published before the SDK, or `workspace:^` resolves to a published `0.2.0` without `SearchTarget.Fact`. Worth a line on the release checklist.

**Also:** both SDK READMEs gained a "Facts (experimental)" section with runnable per-turn examples (your other inline comments), and ADR-0014 is updated — final decisions only, with the rejected designs recorded as rejected alternatives.

Green on the merged tree: Rust workspace (clippy `-D warnings` + fmt), TS 330, vercel-ai-sdk 96, mastra 53, Python 217 + ruff/mypy. Ready for another look when you have a moment.

src/sdk/ts/src/ratel.ts
```diff
@@ -472,7 +566,18 @@ export function ratel(config: RatelConfig = {}): Ratel {
     return { ...base, ...ext } as AdaptedRatel<A>;
```
rstagi: **must-fix: this spread evaluates `get facts()`, so the lazy quarantine never holds for adapter users.**

Object spread does `[[Get]]` on every own enumerable property, so `{ ...base }` invokes the getter defined 20 lines up and builds a `FactCatalog` eagerly. Verified against the built SDK:

```
A: ratel() + tools + recall + modelTools   -> 0 warnings                  (lazy holds)
B: ratel().adaptTo(aiSdk())                -> 1 warning, and `facts` on
                                              the returned view is a plain
                                              VALUE, not a getter
```

So every adapted host gets the experimental warning without touching facts, including both showcase examples (`examples/ai-sdk/src/agent.ts:19`, `examples/mastra/src/agent.ts:15`). This is the PR's central quarantine claim, and the comment at :543 asserts the opposite of what happens.

Fix: attach `facts` with `Object.defineProperty(out, "facts", { get: facts, enumerable: false })` after the spread, or build the returned object without spreading `base`.

Same family, lower stakes: the root object's `facts` getter at :572 is also enumerable, so `JSON.stringify(r)` / `{...r}` / a logger walking the object constructs it too.

src/sdk/ts/src/fact-catalog.ts
```diff
@@ -0,0 +1,321 @@
...
+
+  // Candidates for a grounding pass: always-on facts (pinned) plus the
+  // retrieval-gated facts the query ranks in, deduped by id — a pinned fact
+  // that also ranks appears once, as pinned. Shared by `ground` and
+  // `groundSnapshot` so the two modes can never disagree on the set.
+  private async candidateFacts(query: string, topK: number | undefined): Promise<Fact[]> {
+    const k = clampTopK(topK ?? this.factsTopK, DEFAULT_FACTS_TOP_K);
+    const pinned = this.pinned();
+    const pinnedIds = new Set(pinned.map((f) => f.id));
+    const retrievedHits = await this.searchAsync(query, k, "direct");
```
rstagi: **must-fix: pinned facts consume the retrieval-gated budget, so `topK` does not mean what all three docs say.**

The search runs over both tiers with `k`, then :251 filters pinned back out. Pinned facts occupy ranked slots and are discarded, so the retrieved tier gets `k - (pinned in top-k)`. With the ADR's own motivating example (address, hours, brand voice = 3 pinned, default `topK` 3):

```
snapshot ids:            [p1(always), p2(always), p3(always)]
retrieved (non-pinned):  0
topK: 1 -> [p1]          0 retrieved at every K
```

Identical in the Python `_candidate_facts`. `RatelConfig.factsTopK`, `FactCatalogOptions.factsTopK` and `GroundOptions.topK` all say "max **retrieval-gated** facts". Any host with >= 3 always-on facts has a permanently dead retrieved tier.

Fix: exclude pinned ids from the ranked search, or search `k + pinned.length` and take the first `k` non-pinned.

Root cause for it shipping: `topK` / `factsTopK` / `facts_top_k` have **zero** test coverage in either SDK.

src/sdk/ts/src/grounding.ts
```diff
@@ -0,0 +1,173 @@
...
+export function planInjection(input: PlanInjectionInput): InjectionDecision[] {
+  // Per-message scan: a body counts as present only when contained within a
+  // single message — matching how injections render (one message carries the
+  // whole body). Scanning a joined haystack instead would false-positive when
+  // two unrelated messages happen to end/start with the two halves of a
+  // multi-line body.
+  const previous = input.previouslyInjected;
+
+  return input.candidates.map((candidate): InjectionDecision => {
+    if (candidate.body === "" || input.transcript.some((m) => m.includes(candidate.body))) {
```
rstagi: **Worth a decision: retraction is invisible to the gate.** (Not filing this as a bug, the code matches the spec. I think the spec has the hole.)

Shrinking a body leaves the new text a substring of the stale text still in the transcript, so it reads `fresh` and never re-injects:

```
turn1: cx:never          body "...full refund; same-day is a 50% fee."
edit ->                       "...full refund;"        (fee retracted)
turn2: inject=[] skipped=[cx]
>>> retracted "50% fee" still asserted to the model: true
control (body GROWN):    cx:mutated                    (works fine)
```

Same in Python. I ran three adversarial verifiers at this and all three refuted it, correctly: ADR:83 defines `mutated` as "current body **absent** + differs", and `grounding.test.ts:50` plus the Python twin pin exactly this. So the implementation is faithful.

My disagreement is with the rule, not the code. `ground()` is the persist mode, where injection is the only lever available, and the retraction direction is at least as common as the addition direction ("the shop moves", "the policy changes" is the ADR's own framing). One-line fix, which I ran against all 9 existing vectors:

```ts
// before the presence check: a known-mutated fact whose STALE body is still in
// the window must re-inject, since the new body may be a substring of the old.
if (last !== undefined && last !== c.body && transcript.some(m => m.includes(last)))
  return { id: c.id, inject: true, reason: "mutated" };
```
```
current gate:  1 failing vector    proposed gate: 0 failing vectors
```

Happy either way, but it should be a recorded decision rather than an accident.

docs/adr/0014-facts-and-injection-freshness.md
```diff
@@ -0,0 +1,175 @@
+# 14. Facts: constant grounding content with a content-presence re-injection gate
```
rstagi: **must-fix: ADR number collision.** `0014-adaptive-usage-ranking.md` is already on `main`, along with 0015 and 0016. Two files now open with `# 14.`, and the 11 `ADR-0014` references this PR adds resolve ambiguously.

Next free number is **0017**. AGENTS.md says "next number", and the two merges from main brought the collision in without a renumber.

rstagi (commented):
## Review: experimental facts primitive

**Strong design, faithfully executed, but a few things to fix before merge.** The content-presence gate is a genuinely good idea and the core mechanism works. Findings below are mostly at the edges, plus two real behavioral defects. Everything cited was reproduced by running the code, not read off the diff.

### It builds and runs

| | result |
|---|---|
| `cargo test --workspace` | **334 passed**, 0 failed, 6 ignored |
| `cargo clippy -D warnings` / `cargo fmt --check` | clean |
| TS sdk / ai-sdk / mastra / telemetry | **334 / 96 / 53 / 38** passed |
| `pnpm -r typecheck` / `pnpm -r lint` | clean |
| Python `pytest` / `ruff` / `mypy` | **222 passed**, 1 skipped, clean |

I exercised it end to end in both SDKs across 13 scenarios. Multi-turn dedup holds, `mutated` and `evicted` classify correctly, all five trace events emit with the right payloads, and no fact body reaches telemetry. **TS/Python parity: 13/13 scenarios byte-identical.** The mirroring is faithful.

### Blocking

1. **`ratel.ts:566`** the spread evaluates `get facts()`, so every `adaptTo()` host eagerly builds the experimental catalog. Falsifies the quarantine claim; both showcase examples hit it.
2. **`fact-catalog.ts:248`** pinned facts eat the `topK` budget, zeroing the retrieved tier at the ADR's own 3-fact example.
3. **ADR renumber to 0017**, plus three stale claims in it (the empty-catalog gate, `recall()` consulting the catalog, the recall/adapter seam).

Plus one design question inline on `grounding.ts:111` about retraction, which is correct against the spec but I think the spec has a hole. Fix included and validated against all existing vectors.

### Not in-lineable

**`src/telemetry/CONVENTIONS.md:186`** still pins `ratel.search.target` to the two-value enum `tool | skill`, but all three telemetry packages now emit `fact`. The inline docs in `telemetry/core` and `telemetry/ts` were updated in this PR; the wire-contract doc was missed. That file is the vendor-facing contract, so it should move in the same commit.

### What is clean, explicitly

Not padding the list, several things are done well and I checked them specifically:

- **The compatibility claim holds at runtime.** `capabilities.ts` and `capabilities.py` are byte-untouched, `recall()`'s body is unchanged, and `recall()` returns exactly `{tools, skills}` with no `fact_search` even with facts registered, in both SDKs. The subtractive "one path into context" commit did its job.
- **`FactRegistry` is a clean structural twin of `SkillRegistry`**, minus the usage/intent-graph arm as intended. `fact_indexing.rs` correctly excludes `body` and `metadata`, with tests proving it.
- **Bindings are solid.** GIL released where the skill path releases it, poisoned locks handled, no panics across FFI, all seven `Fact` fields round-trip in both directions in both SDKs.
- **Telemetry is correct** and consistent across all three languages.
- **Python's quarantine is stronger than TS's**, since there is no `ratel()` object to leak through.

### Should this be a stack?

4652 / 24 across 36 files clears the bar, and there is a real seam: **TS and Python share zero code, only the Rust core.** The two Python-only defects (the `Sequence[str]` hole, the forked clamp) are the signature of reviewer fatigue setting in on 1832 lines of mirror after 1669 lines of original.

| # | title | base | paths | LOC |
|---|---|---|---|---|
| 1 | `feat(telemetry): add SearchTarget.Fact` | `main` | `src/telemetry/**` | ~14 |
| 2 | `feat(core): Fact/PinMode/FactRegistry + fact trace events` | #1 | `src/core/**`, ADR, root+core README | ~1250 |
| 3 | `feat(sdk-ts): experimental facts + grounding gate` | #2 | `src/sdk/ts/**` | ~1670 |
| 4 | `feat(sdk-py): experimental facts + grounding gate` | #3 | `src/sdk/python/**` | ~1830 |

Linear, so 4 sits on 3 even though the two SDKs share no code. The order between them is arbitrary. The win is that each lands green on its own and reviews as one coherent unit, not that they go in parallel.

Entirely your call though. The bugs need fixing either way, and the ADR plus core are cohesive enough that landing as one PR after fixes is defensible. Branch is current with `main` and the runtime compatibility claims check out, so this is fixable in place rather than a rework.

### Smaller things (inline)

`FactCatalog._rebuild` in Python has no caller. `FactHit` claims `SkillHit` parity but lacks the `rank`/`fused` its docs point you at. Python's `ground_snapshot` docstring cites `appendRecall`/`prepareStep`, which do not exist in the Python SDK. `native/README.md` still lists only tool and skill registries.

<sub>Reviewed against `b640fecf`. Repro details in the inline comments.</sub>

docs/adr/0014-facts-and-injection-freshness.md
```diff
@@ -0,0 +1,175 @@
...
+Facts ship **experimental** — a new, unproven API we want to trial in the open before committing to
+stability. The whole surface (catalog, registry, planner, `Pin`, the grounding types) is reachable
+only through an `experimental` namespace (`experimental.FactCatalog` in TS via `export * as
+experimental`; `ratel_ai.experimental` in Python), never the root export, so any dependence on it is
+explicit at the import site. Constructing a `FactCatalog` logs a one-time warning (silence:
+`RATEL_EXPERIMENTAL_SILENCE`). The `ratel()` touchpoints that can't move off the stable object
+(`r.facts`, `r.ground`, `r.groundSnapshot`, `RatelConfig.factsTopK`) are tagged "⚠️ Experimental" in their docs, and **`r.facts` is lazy**: the catalog is constructed
+on first access, so a host that never touches facts never builds one and never sees the
+experimental warning. `recall()` likewise consults the catalog only if it already exists. Two
+further boundaries keep the experiment off the stable path: the fact search is skipped on an empty
```
rstagi: This gate no longer exists. It went out with the recall facts bucket, and `ground` / `groundSnapshot` search unconditionally:

```
TS events on an EMPTY fact catalog: [ 'fact_search', 'fact_search' ]
```

Not a behavior problem now (a host with no facts never calls `ground`), but the ADR states it as a live quarantine boundary. Drop it or restate it as removed.

docs/adr/0014-facts-and-injection-freshness.md
```diff
@@ -0,0 +1,175 @@
...
+
+Facts ship **experimental** — a new, unproven API we want to trial in the open before committing to
+stability. The whole surface (catalog, registry, planner, `Pin`, the grounding types) is reachable
+only through an `experimental` namespace (`experimental.FactCatalog` in TS via `export * as
+experimental`; `ratel_ai.experimental` in Python), never the root export, so any dependence on it is
+explicit at the import site. Constructing a `FactCatalog` logs a one-time warning (silence:
+`RATEL_EXPERIMENTAL_SILENCE`). The `ratel()` touchpoints that can't move off the stable object
+(`r.facts`, `r.ground`, `r.groundSnapshot`, `RatelConfig.factsTopK`) are tagged "⚠️ Experimental" in their docs, and **`r.facts` is lazy**: the catalog is constructed
+on first access, so a host that never touches facts never builds one and never sees the
+experimental warning. `recall()` likewise consults the catalog only if it already exists. Two
```
rstagi: Also stale: `recall()` does not consult the catalog under any condition. It returns `{tools, skills}` and emits no `fact_search` even with facts registered, which I verified. The "only if it already exists" wording is residue from the pre-removal revision.

docs/adr/0014-facts-and-injection-freshness.md
```diff
@@ -0,0 +1,175 @@
...
+
+Date: 2026-07-27
+
+## Status
+
+Accepted
+
+Builds on first-class skills (ADR-0005), the selectable retrieval engines (ADR-0004,
+ADR-0011), the two telemetry streams (ADR-0007), and the framework-adapter SPI (ADR-0013):
+facts reuse the skill registry pattern wholesale and surface through the same `recall`/adapter
```
rstagi: Stale preamble: facts no longer "surface through the same `recall`/adapter seam". This contradicts the Decision section below, which records removing exactly that seam.

src/sdk/ts/README.md
```diff
@@ -134,6 +134,61 @@ the un-adapted core throws an error pointing at the adapter package to install.
...
+  },
+]);
+```
+
+Then pick **one** of two injection modes per turn — the same persist-vs-per-call split as `appendRecall` vs `prepareStep` for recall.
+
+**`ground()` — persist into your stored history.** Returns only the facts not already present; render each `body` verbatim and keep it in the messages you save:
+
+```ts
+const { inject } = await r.ground(userText, messages.map((m) => m.content));
```
rstagi: **This snippet silently defeats the gate on Vercel AI SDK messages,** the framework the SDK leads with.

`ModelMessage.content` is `string | Array<Part>`, which this repo's own adapter handles both ways (`aisdk.ts:345` and `:381`). When it is a part array, `planInjection` calls `Array.prototype.includes`, which is element equality, not substring, so it never matches:

```
turn 1: never
turn 2: evicted
turn 3: evicted     ... forever
```

No error, no warning, and the feature's entire value proposition is inverted into per-turn token waste. TS only catches it if the host has precisely typed `ModelMessage[]`; `any` / `UIMessage` / JSON-loaded history all pass.

Worth either exporting a text-extraction helper for grounding, or at minimum flagging the hazard here and in the ADR's known limits.

src/sdk/python/ratel_ai/grounding.py
```diff
@@ -0,0 +1,180 @@
...
+        One `InjectionDecision` per candidate, in the same order.
+    """
+    # Per-message scan: a body counts as present only when contained within a
+    # single message — matching how injections render (one message carries the
+    # whole body). Scanning a joined haystack instead would false-positive when
+    # two unrelated messages happen to end/start with the two halves of a
+    # multi-line body.
+    decisions: list[InjectionDecision] = []
+    for candidate in candidates:
+        if candidate.body == "" or any(candidate.body in message for message in transcript):
```
rstagi: `transcript: Sequence[str]` accepts a bare `str`, since `str` is itself a `Sequence[str]`. It then iterates characters and the gate is permanently defeated:

```
c.ground("x", "the transcript contains AAA verbatim right here")
->  inject=[('a', 'never')]  skipped=[]     # expected: skipped
```

mypy will not catch this. TS's `readonly string[]` rejects it at compile time, so this is a Python-only hole. Suggest an `isinstance(transcript, str)` guard raising `TypeError`.

src/core/src/fact_registry.rs
```diff
@@ -0,0 +1,643 @@
...
+    /// Replacing invalidates the old id's cached embedding; the corpus never
+    /// holds a duplicate.
+    pub fn register(&mut self, fact: Fact) {
+        let fact_id = fact.id.clone();
+        if self.facts.insert(fact_id.clone(), fact).is_some() {
+            // Replaced an existing id: drop its stale embedding.
+            self.dense.invalidate(&fact_id);
+        }
+        // The corpus changed: the cached BM25 index is stale.
+        self.bm25.invalidate();
```
rstagi, line 115: This `invalidate()` is completely untested. Every test in the module registers *before* searching, so the BM25 cache is always cold at mutation time. I removed the line and all 334 Rust tests stayed green.

`skill_registry.rs:929` has the regression test for exactly this (`bm25_cache_is_warmed_by_search_and_dropped_by_every_mutator`), added when the index cache landed in #144. Worth porting: a stale index means `ground()` ranks against the old corpus, so an edited fact never surfaces in the retrieved tier.

src/sdk/ts/src/registry.ts
```diff
@@ -483,3 +486,112 @@ export class SkillRegistry {
...
+  }
+
+  /**
+   * Register one fact or a batch, replacing any existing id in place — see
+   * {@link ToolRegistry.register} for the embed-inside contract. An unknown
+   * `pin` value throws here, synchronously, before the batch is indexed.
+   *
+   * @param item - A single {@link Fact} or a readonly array of them.
+   */
+  async register(item: Fact | readonly Fact[]): Promise<void> {
```
rstagi, line 520: This public export bypasses the id validation that `FactCatalog.register` enforces. `FACT_ID_PATTERN` is checked at the catalog boundary only:

```
ACCEPTED by experimental.FactRegistry: {"id":"has space"}
ACCEPTED by experimental.FactRegistry: {"id":"id\nwith\nnewlines"}
ACCEPTED by experimental.FactRegistry: {"id":""}
rejected: pin "sometimes"     (that one is caught natively)
```

ADR known-limit 4 justifies the constraint by "ids ride in trace events and structured injection payloads", which applies equally to both entry points.

Minor, same area: the JSDoc at :515 promises the bad-`pin` throw happens "synchronously", but the method is `async`, so it surfaces as a rejected promise.

src/sdk/ts/src/fact-catalog.ts
```diff
@@ -0,0 +1,321 @@
...
+export class FactCatalog {
+  private readonly registry: FactRegistry;
+  private readonly facts = new Map<string, Fact>();
+  private readonly method: SearchMethod;
+  private readonly factsTopK?: number;
+  // Session bookkeeping for the freshness gate: the body this catalog last
+  // injected per fact id via `ground`. Lets an absent body be classified as
+  // `evicted` (same body, gone from the window) or `mutated` (body has since
+  // changed) instead of `never`. The transcript itself carries the rest.
+  private readonly injectedBodies = new Map<string, string>();
```
rstagi, line 70: Worth documenting that this map is scoped to the catalog, not to a conversation.

On a shared `ratel()` (which the README's module-scope example encourages for servers), the injection *decisions* stay correct, since presence is computed from the per-call `transcript`. Nice design. But the `reason` labels cross-contaminate:

```
userA turn1: [ 'never' ]
userB turn1: [ 'evicted' ]     # B's first ever injection, labelled from A's state
```

That degrades the `fact_inject` telemetry the READMEs sell as "the skip rate, the tokens you saved, is measurable" in exactly the multi-tenant shape the docs suggest.

src/sdk/ts/native/src/lib.rs
```diff
@@ -844,6 +844,376 @@ impl ToolRegistry {
...
+    pub tags: Option<Vec<String>>,
+    /// Free-form, non-indexed context for higher layers — e.g.
+    /// `{ stacks: ["react"] }` for the push ranker to boost by project context.
+    pub metadata: Option<HashMap<String, Vec<String>>>,
+    /// The content injected into the context on grounding — the payload, never
+    /// indexed for ranking. Optional (defaults to `""`).
+    pub body: Option<String>,
+    /// `"always"` (always-on) or `"retrieved"` (retrieval-gated). Optional
+    /// (defaults to `"retrieved"`); an unknown value throws at registration.
+    pub pin: Option<String>,
```
rstagi, line 971: `Option<String>` generates `pin?: string` in the public type, so the `Pin` const object buys no type safety at the call site:

```ts
await r.facts.register({ id: "x", pin: "sometimes" })  // typechecks, throws at runtime
```

The ADR presents the contract as `pin: "always" | "retrieved"`. Suggest declaring the SDK's own `Fact` with `pin?: Pin` rather than re-exporting the napi type verbatim from `fact-catalog.ts:19`.

src/sdk/ts/src/fact-catalog.test.ts
```diff
@@ -0,0 +1,361 @@
...
+    const transcript = first.inject.map((i) => i.body);
+    await r.ground("hi", transcript);
+    const events = r.facts.drainTraceEvents() as Array<{ type: string; fact_id: string }>;
+    const types = events.map((e) => e.type);
+    expect(types).toContain("fact_inject");
+    expect(types).toContain("fact_inject_skip");
+  });
+});
+
+describe("experimental quarantine on the stable path", () => {
```
rstagi, line 427: This suite is good, and `grounding.test.ts` is genuinely strong (all three reasons pinned by exact equality, plus the per-message-scan false-positive guard). Gaps I mutation-tested, each of which leaves both SDK suites fully green:

- **the `adaptTo` quarantine** is untested, which is why the eager-construction bug shipped. These cases only cover the root object, and the adapted-view test at :341 touches `view.facts` deliberately.
- **`topK` / `factsTopK` / `facts_top_k`**: zero coverage in either SDK. Hardcoding the default changes nothing.
- **`fact_inject`'s `reason`** is never asserted. Hardcoding it to `"never"` passes.
- **the pinned-vs-retrieved dedup** at `fact-catalog.ts:251`: removing it passes.
- **the `target="fact"` OTel span**: stripping `traceSearch` passes.
- **`RATEL_EXPERIMENTAL_SILENCE`**: tested in Python, not TS.
- **3+ turn sequences**: everything here is at most two turns, so nothing pins that `injectedBodies` refreshes after a `mutated` re-inject.

Also asymmetric: Python added two quarantine tests to `test_index.py`; TS's `index.test.ts` has no `experimental` or `Fact` reference at all.

src/sdk/python/ratel_ai/fact_catalog.py
```diff
@@ -0,0 +1,731 @@
...
+
+    A `str` enum, so `Pin.ALWAYS` is interchangeable with the wire string
+    `"always"` — pass either to `Fact(pin=...)`.
+    """
+
+    ALWAYS = "always"
+    RETRIEVED = "retrieved"
+
+
+def _clamp_facts_top_k(value: int | None) -> int:
```
rstagi, line 103: nit: this forks `capabilities._clamp_top_k` (identical body, including the `bool` guard) and redefines `_MAX_TOP_K = 50`. The TS twin correctly imports the shared `clampTopK` from `./capabilities.js`. Two copies of the cap will drift.

kayraucklnc:
All 16 comments from the second pass verified and addressed in `8e3780a`. Every one was correct — I reproduced each before fixing, and there were no false positives. Thank you, this was a genuinely better review than the checks I was running.

### The three behavior bugs

**`adaptTo` built the catalog eagerly.** Confirmed exactly as described: `ratel()` + recall + modelTools → 0 warnings, `ratel().adaptTo(…)` → 1, with `facts` a plain value on the returned view. Painful one, because that getter *was* my fix for your first-round must-fix — I verified it on the root object and never enumerated the other path. Now attached via `Object.defineProperty(…, { enumerable: false })` in a `withLazyFacts` helper used for both the base and the spread result, so spread / `Object.assign` / `JSON.stringify` / a key-walking logger all stay quiet.

**Pinned facts ate the `topK` budget.** Your worst case reproduces: 3 pinned + `topK: 3` with a pinned-favouring query → `["p1","p2","p3"]`, zero retrieved. Now searches `k + |pinned|` and truncates to `k` *after* the filter, in both SDKs. The ADR states the `topK`-budgets-the-retrieved-tier-only semantics explicitly now.

**Python's bare-`str` transcript.** Reproduced verbatim. Guarded with a `TypeError` in `plan_injection` and at the `ground()` boundary, with the reason in the message.

### The retraction hole

You're right and I took your rule. The ADR's own framing is "the shop moves, the policy changes" — retraction is at least as common as addition, and `ground()` has no other lever. It's recorded as a decision (with the honest caveat: an append-only transcript can't drop the stale copy, so the correction goes last rather than replacing it) and the presence-only rule is now in Rejected alternatives.

### The rest

ADR renumbered **0014 → 0017** (0014 arrived from main via #144 — I had it in my own directory listings and merged past it twice, purely my miss). Three stale passages dropped: the `recall`/adapter-seam preamble, "recall consults the catalog only if it already exists", and the empty-catalog search skip, which genuinely no longer exists. Id validation moved into `FactRegistry` so both entry points enforce it. The SDK now declares its own `Fact` with `pin?: Pin` rather than re-exporting napi's `pin?: string`. README snippets extract message text before grounding, with the `string | Array<Part>` hazard called out and added to the ADR's known limits. Python clamp deduped onto the shared one. Catalog-scoped injection state documented rather than fixed — the *decisions* are correct and only the `reason` labels cross-contaminate, so a known-limit entry felt proportionate under `experimental`; say the word if you'd rather it were per-conversation.

### Tests

Your mutation-testing list was the most useful part, so I closed it that way: `adaptTo` quarantine, `topK`/`factsTopK`, `fact_inject` reason fidelity, pinned/retrieved dedup, `target="fact"` span, `RATEL_EXPERIMENTAL_SILENCE` in TS, a 3-turn sequence pinning the injected-body refresh, TS/Python export-surface parity, and the BM25 cache-invalidation regression ported from `skill_registry.rs:929`. **Each one mutation-verified** — I reverted every fix individually and confirmed the matching test goes red before restoring.

Green across the matrix: Rust 336 + clippy/fmt, TS 359 + typecheck/biome/typedoc, Python 240 + ruff/mypy, vercel-ai-sdk 96, mastra 53.

Worth another look when you have time.

rstagi (approved):
lgtm 🚀

PR #132, opened 2026-07-24
test(telemetry-otlp): prove Langfuse + AI SDK 7 + host-provider coexistence

rstagi:
Adds a real-dependency coexistence test suite (`src/telemetry/ts-otlp/src/coexistence.test.ts`) for the TS telemetry DX northstar (RS-43, Phase 4), proving Ratel telemetry shares one OpenTelemetry pipeline with Langfuse and the AI SDK 7 `@ai-sdk/otel` integration. It covers guest mode (host provider + `LangfuseSpanProcessor` + `ratelSpanProcessor` with the default signal filter), owner mode (`startTelemetry` + `registerTelemetry(new OpenTelemetry())`), and per-destination `ratel.*`-only filtering — each backend observed at its own OTLP exporter boundary, no network — and adds `@langfuse/otel`, `@ai-sdk/otel`, and `ai` as ts-otlp devDeps. Finding for the docs phase (RS-44): AI SDK 7's `@ai-sdk/otel` integration emits `gen_ai.*`-tagged `chat`/`step`/`invoke_agent` spans, so `ratelSignalFilter` keeps them all; the "drops ai.* noise" caveat applies only to the legacy `experimental_telemetry` path. Full TS workspace green (ts-otlp 33, ts 54, sdk 265, adapters 71+53).

PR #137, opened 2026-07-26
test(vercel-ai-sdk): prove Langfuse + AI SDK 7 + host-provider coexistence

rstagi:
Adds `src/adapters/ts-vercel-ai-sdk/src/coexistence.test.ts` (7 tests): one host-owned provider fans the SDK's `ratel.*` spans and the AI SDK's `gen_ai.*` spans to several destinations at once, including the real `LangfuseSpanProcessor` captured at its own exporter, with per-destination filtering shown via Langfuse's `shouldExportSpan` beside a hand-written predicate. No network and no Ratel Cloud import, since `@ratel-ai/cloud` ships last and the OSS path is proven against third-party backends first; `@langfuse/otel` is added as a devDep and the lock gains no second `ai` copy, no `@mastra/core` peer flip, and `@opentelemetry/api` stays a single 1.9.1.

**Finding for the docs phase (RS-44):** Langfuse's default `shouldExportSpan` keeps a span only when it carries a `gen_ai.*` attribute or comes from a scope on its known-instrumentor list, and `@ratel-ai/sdk` is on neither, so a stock `new LangfuseSpanProcessor()` silently drops `ratel.search`, `ratel.skill.load`, `ratel.upstream.register` and `ratel.auth.flow` while `execute_tool` survives on its `gen_ai.tool.name`. Both that behaviour and the one-line host predicate that opts Ratel back in are pinned by tests.

**Second finding:** `otel.scope.name` identifies the emitter, not the span name or attribute prefix, since both emitters produce an `execute_tool <id>` span and the SDK's carries `gen_ai.*` attributes without the integration's `enrichSpan` attributes; the suite keys on scope throughout and pins both names.

Verified: 95/95 on ai@7.0.37, ai@5.0.0 and ai@6.0.0 compat rows typecheck and skip the ai@7 block cleanly, `pnpm -r test` green (437 tests), lint and typecheck clean, and every load-bearing assertion mutation-tested.

Plan: doc `15Pxf82c` phase C (RS-43). Supersedes #132, which was closed because its only file lived inside the `src/telemetry/ts-otlp` package that #135 deleted.


rstagi:
## RS-44 (Phase D) — the v2 telemetry docs, bundled here

Rides with RS-43 on this branch per direction, rather than the plan's suggested `docs/telemetry-v2-examples`. Commit e92c6ca.

**`src/telemetry/README.md` is now the single home for host wiring** — span/EventRecord inventory, the emission-vs-delivery model, and the three OSS-generic scenarios. The package READMEs keep what is local to them and link there, because the alternative had already failed: the bad `new LangfuseSpanProcessor()` snippet existed in **three** places (`src/sdk/ts/README.md`, `src/adapters/ts-vercel-ai-sdk/README.md`, and the shipped TSDoc in `src/adapters/ts-vercel-ai-sdk/src/otel.ts`, which surfaces on hover in every consumer's editor) and was wrong in all three. All corrected, keyed on scope.

`src/adapters/ts-mastra/README.md` gets its first telemetry section. `llms.txt` gets a telemetry pitfall. ADR-0007's interop consequence gains the delivery caveat. `telemetry-wiring.type-test.ts` pins the predicate's argument shapes, which the behavioural suite does not cover.

### Two things the plan's appendix got wrong, both caught before shipping

- **Appendix example 3 (Mastra) does not compile.** `Observability` is not exported from `@mastra/core`; it lives in a separate `@mastra/observability` package. Published as written it is a `TS2305` plus `undefined is not a constructor`. Four more corrections came with it: passing a plain config object instead of an instance silently installs `NoOpObservability` rather than throwing; the OTLP exporter is a *peer* of `@mastra/otel-exporter` and a missing one degrades to zero spans with only a debug log; `endpoint` is a base URL; and the telemetry stack floors at `@mastra/core >= 1.16`, five minors above the adapter's own `>= 1.11.0`.
- **Mastra 1.51 emits nothing to OTel un-instrumented.** The prior working assumption that Mastra 1.x auto-emits `gen_ai.*` is dead — those strings in its bundle are a vendored AI SDK copy gated off by default, and the legacy `telemetry` constructor option is gone from `Config`. Probed on both the bare-`Agent` and `Mastra`-instance paths: zero spans. Its pipeline is also fully private (host delegate identity preserved, zero leaked spans), so there is no NodeSDK conflict to warn about — but host processors never see Mastra spans either.

### Verification

Every fenced snippet compiles, not just the new ones — 21 `tsc` runs across the six touched files. The AI SDK wiring compiles against **npm-published** `@ratel-ai/vercel-ai-sdk@0.2.0` + `@ratel-ai/sdk@0.5.3` in a consumer with no repo on its resolution path, and `/otel` resolves there at runtime. The Mastra snippet was typechecked *and* run end-to-end against a local OTLP catcher (5 protobuf batches to `/v1/traces`, auth header intact) on `@mastra/core@1.51.0` + `@mastra/observability@1.16.2` + `@mastra/otel-exporter@1.3.5`. Locally: `pnpm -r typecheck`, `pnpm -r lint`, adapter suite 95/95.

An adversarial audit pass over the diff then killed six of my own claims, worth recording since two were live regressions:

- I had written that `./otel` "is not in `0.1.0` and ships with the next release". **`0.2.0` shipped it ~9h earlier** and is `latest`; the branch is one commit behind `6496da4`. Replaced with a plain `>= 0.2.0` floor.
- My `llms.txt` edit to the adapter-availability line **regressed a better version already on main**. Reverted to base so the merge takes main's; the only `llms.txt` change left is the additive telemetry pitfall.
- "On `ai@5`/`ai@6` there is no integration seam" is false for `ai@6 >= 6.0.108`, which has the earlier `registerTelemetryIntegration`.
- The `execute_tool <tool name>` row was missing `ratel.tool.args_size_bytes` (always set) and `ratel.upstream.server` (for `<server>__<tool>` ids) — it is the only *mixed* span, not a pure `gen_ai.*` one.
- My Mastra `endpoint` bullet claimed a full traces URL misroutes logs. It does not: `getSignalEndpoint` strips either signal suffix per signal, so both resolve correctly.
- "Typechecks against nothing useful" for `BatchLogRecordProcessor` was wrong — `tsc` does catch it (`exporter` is required); the silent drop is the JavaScript case only.

### Follow-ups this surfaced, deliberately not in scope here

- [ ] `CONVENTIONS.md` lists `ratel.origin` on the `execute_tool` span and `conformance/fixtures.json` expects it there, but neither SDK emits it — both set it on `ratel.search` only. Spec/impl drift; changing the wire contract is not a docs PR's call. Noted as a known gap in the new README.
- [ ] `gen_ai.tool.call.id` is declared in the vocabulary and never emitted anywhere.
- [ ] This branch is 1 commit behind `origin/main` (`6496da4`, the 0.2.0 release). Worth rebasing before merge so the branch's `package.json` is not `0.1.0` against a README asserting `>= 0.2.0`.
- [ ] `src/adapters/ts-vercel-ai-sdk/README.md`'s Install section still says the adapter "is currently a release candidate" and to install the `@rc` pair. `latest` is `0.2.0` GA, and `@rc` resolves to `0.1.0-rc.5`, which has **no** `./otel` — so a reader following Install cannot use the telemetry section. Pre-existing on main, left for its owner.
- [ ] The repo has **no** markdown-snippet verification at all, and `@opentelemetry/sdk-node` is absent from the lockfile despite `NodeSDK` appearing in two README snippets. The cheapest durable seam is the `ai-sdk-compat` v7 leg in `ts.yml`, which already has a warm consumer dir.
- [ ] `docs-rebuild.yml` watches only `src/sdk/ts/README.md` and `src/sdk/python/README.md`, so none of these telemetry edits reach docs.ratel.sh.
- [ ] `llms.txt`'s "don't expose every catalog tool" pitfall shows `catalog.tools`, which is `private readonly` — the ❌ example fails with `TS2341`, not the failure its comment describes.
- [ ] `RELEASING.md:138` — `@ratel-ai/telemetry-otlp@*` is still live and undeprecated on npm.
- [ ] `examples/telemetry-{ts,python}` READMEs advertise `https://cloud.ratel.sh/v1/traces`, while the locked Cloud contract is `/api/v1/traces`. Phase E's (RS-49) to settle.

Full verified-facts record for whoever picks up RS-49: `.context/rs44-verified-facts.md` (gitignored).

PR #138, opened 2026-07-27
feat(sdk): paginated MCP tools/list in TS sdk and Python sdk

faustoq:
### Summary 

> Follow MCP `tools/list` cursors until exhaustion in both SDKs before reporting registration complete.

**Problem:** registerMcpServer / register_mcp_server only handled the first tools/list page. On paginated upstream servers, tools on later pages were never registered. `upstream_register.tool_count` reflected that first page only, not the full aggregated listing.

TS (`@ratel-ai/sdk`): Loop on nextCursor until absent (empty string is still a valid cursor), aggregate tools, single catalog.register, 64-page cap + repeated-cursor errors. List/register failures close the SDK-owned Client. Pagination + failure cases covered in mcp.test.ts (including empty page, later-page list failure, embed failure + close).

Python (`ratel-ai`): Same listing semantics via _list_all_mcp_tools (PaginatedRequestParams / fake-compatible params.cursor). Caller-owned ClientSession unchanged—no auto on_close on failure. Tests in test_mcp.py mirror TS pagination behavior.

--

### Review notes
Duplicate tool names across pages: ids listed in order; catalog batch keeps last row on invoke (documented on handles).
Python test_register_mcp_server_requires_mcp_when_absent skips when the mcp extra is installed (same as CI).

### DIFF
--- a/src/sdk/python/ratel_ai/mcp.py
+++ b/src/sdk/python/ratel_ai/mcp.py
@@ -58,9 +58,9 @@ def _require_mcp() -> None:
 
 
 def _next_cursor(result: Any) -> str | None:
-    nc = getattr(result, "next_cursor", None)
+    nc = getattr(result, "nextCursor", None)
     if nc is None:
-        nc = getattr(result, "nextCursor", None)
+        nc = getattr(result, "next_cursor", None)
     return nc
 
 
--- a/src/sdk/python/ratel_ai/mcp.py
+++ b/src/sdk/python/ratel_ai/mcp.py
@@ -1,9 +1,9 @@
 """Upstream MCP ingestion — the Python mirror of `src/sdk/ts/src/mcp.ts`.
 
-`register_mcp_server` lists an upstream MCP server's tools, registers each into a
-`ToolCatalog` under a namespaced id (`<server>__<tool>`), and wires each executor
-to the upstream `call_tool`. It emits the same `upstream_*` trace **event types**
-the TS SDK emits (ADR-0007).
+`register_mcp_server` lists every paginated upstream MCP `tools/list` page (no live
+refresh), registers each tool into a `ToolCatalog` under a namespaced id
+(`<server>__<tool>`), and wires each executor to the upstream `call_tool`. It emits
+the same `upstream_*` trace **event types** the TS SDK emits (ADR-0007).
 
 The `mcp` package is an optional dependency (`pip install ratel-ai[mcp]`) and is
 imported lazily, so the base SDK installs without it.
@@ -27,10 +27,24 @@ from __future__ import annotations
 import time
 from collections.abc import Awaitable
 from dataclasses import dataclass
-from typing import Any, Callable
+from typing import Any, Callable, Literal
 
 from .catalog import ExecutableTool, ToolCatalog
 from .telemetry import trace_upstream_register
+
+_MCP_LIST_MAX_PAGES = 64
+
+McpToolsListErrorCode = Literal["RepeatedCursor", "PaginationExceeded"]
+
+
+class McpToolsListError(Exception):
+    """Raised when MCP `tools/list` pagination is invalid or exceeds the page cap."""
+
+    def __init__(self, message: str, code: McpToolsListErrorCode) -> None:
+        """Set the error message and pagination failure code."""
+        super().__init__(message)
+        self.code = code
+        self.name = "McpToolsListError"
 
 
 def _require_mcp() -> None:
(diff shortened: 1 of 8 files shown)


kayraucklnc (commented):
I checked `register_mcp_server` against a real `mcp.ClientSession` (real in-memory transport, not the fakes) to sanity check the Python side. It works on `mcp==1.27.2`, but that probe turned up one compat problem worth fixing before merge, plus a few smaller notes inline.

src/sdk/python/ratel_ai/mcp.py
```diff
@@ -43,13 +57,86 @@ def _require_mcp() -> None:
...
+    except ImportError:
+        from types import SimpleNamespace
+
+        return SimpleNamespace(cursor=cursor)
+
+
+async def _fetch_tools_page(session: Any, cursor: str | None) -> Any:
+    if cursor is None:
+        return await session.list_tools()
+    return await session.list_tools(params=_paginated_list_params(cursor))
```
kayraucklnc: `params=` is keyword-only and only landed in `mcp` 1.18, but `pyproject.toml` declares `mcp>=1.0`. I checked the real signature across releases:

| version | `ClientSession.list_tools` |
| --- | --- |
| 1.0.0 | `(self)` |
| 1.9 to 1.16 | `(self, cursor: str \| None = None)` |
| 1.18+ | `(self, cursor=None, *, params=None)` |

On 1.16.0 this raises `TypeError: ClientSession.list_tools() got an unexpected keyword argument 'params'`. So for anyone inside the supported range but below 1.18, a paginating server goes from "silently truncated at page 1" to a hard crash mid-registration, which makes it a regression rather than just a missing feature.

CI won't catch it: the Python workflow installs bare `mcp`, which resolves to latest, and there's no lockfile.

Simplest fix is `cursor=` instead, supported since 1.9 and still accepted on 1.27. Alternatives: try `params=` and fall back on `TypeError`, or bump the floor to `mcp>=1.18` and pin the CI install so the floor is actually exercised.

Side effect of fixing it either way: `_paginated_list_params` can go entirely. Its `except ImportError` branch is unreachable in production (`_require_mcp()` already ran), and in the one case it would fire (mcp < 1.12, no `PaginatedRequestParams`) you hit the same `TypeError` a line later anyway. Right now it only exists so the fakes work without the extra installed.

src/sdk/python/tests/test_mcp.py
```diff
@@ -22,15 +38,16 @@ def __init__(self, name, description, input_schema):
 
 
 class _FakeListResult:
-    def __init__(self, tools):
+    def __init__(self, tools, *, next_cursor: str | None = None):
         self.tools = tools
+        self.next_cursor = next_cursor
```
kayraucklnc: The real `mcp.types.ListToolsResult` field is `nextCursor` (camelCase on every version from 1.0 through 1.27). `_next_cursor` tries `next_cursor` first, so all ten new pagination tests take the snake_case branch and the branch that runs against real servers never executes.

It does work, I confirmed it with a real `ClientSession`, but nothing here proves that: swap the two `getattr` lines in `_next_cursor` and the whole suite stays green.

Would suggest either renaming this to `nextCursor` so the fakes match the wire types, or adding one test against a real session via `mcp.shared.memory.create_connected_server_and_client_session`. The second option would also have caught the `params=` issue above, which is a decent argument for it.

src/sdk/ts/src/mcp.ts
```diff
@@ -1,9 +1,96 @@
...
+import type { Tool } from "@modelcontextprotocol/sdk/types.js";
 
 import type { ExecutableTool, ToolCatalog } from "./catalog.js";
 import { traceUpstreamRegister } from "./telemetry.js";
 
+const MCP_LIST_MAX_PAGES = 64;
+
+type McpToolsListErrorCode = "RepeatedCursor" | "PaginationExceeded";
+
+class McpToolsListError extends Error {
```
kayraucklnc: Not exported here, and not re-exported from `index.ts`. Same story on the Python side: `mcp.py` defines it but `ratel_ai/__init__.py` only exports `McpServerHandle` and `register_mcp_server`, while the `register_mcp_server` docstring documents `Raises: McpToolsListError`.

So callers can't `instanceof` it or `except` it, they can only string-match `err.code`, which is what the tests end up doing. Either export it from both public surfaces or drop it from the documented contract.

src/sdk/ts/src/mcp.ts
```diff
@@ -82,64 +172,33 @@ export async function registerMcpServer(
...
-            throw err;
-          }
-        },
+      const tools = await listAllMcpTools(client);
+      reportToolCount(tools.length);
+      catalog.recordEvent({
+        type: "upstream_register",
+        server: name,
+        transport: transportLabel,
+        tool_count: tools.length,
```
kayraucklnc, line 197: Small note on the PR description rather than the code: this is still the raw listing length, so with duplicate names across pages it reports 2 while the catalog holds one row (the dup test asserts `toolIds` is `["demo__dup", "demo__dup"]`).

The tradeoff is deliberate and documented on the handle, so no change needed. Just flagging that "upstream_register.tool_count was wrong" is slightly overstated: it's now right for the common case, but it's a listing count, not a catalog count.

src/sdk/python/ratel_ai/mcp.py
faustoq: Good catch! Switched paginated calls to `list_tools(cursor)` and dropped `_paginated_list_params`, so we stay compatible with mcp>=1.0. The in-memory ClientSession test should guard this going forward.

src/sdk/python/tests/test_mcp.py
faustoq: fakes now use `nextCursor` and `_next_cursor` prefers camelCase. The in-memory session coverage has been added too (see the thread above)

src/sdk/ts/src/mcp.ts
faustoq: that's right! PR description updated to reflect this

src/sdk/ts/src/mcp.ts
faustoq: Done. Exported `McpToolsListError` from both packages entrypoints

faustoq:
Thanks for the review @kayraucklnc ! I applied the requested fixes. 

Also a small `telemetry.py` [change](https://github.com/ratel-ai/ratel/pull/138/changes/89baa976a8ab868a7ac21cda7dd902a704205285) was required so OTel tool-result capture stays stable on mcp 2.x (CI was failing once CallToolResult started including fields like resultType in serialized spans)

rstagi (approved):

kayraucklnc (approved):
LGTM

PR #139, opened 2026-07-27
feat(sdk): whole-catalog skill reload via replaceAll

alesdi:
## Why

A catalog source — the cloud client first — reloads the full skill catalog and hands it to the SDK. There was no way to apply it: `SkillCatalog` was append-only (`register` replaces an id in place, nothing removes one), so a reload could only ever grow the corpus. A skill deleted upstream stayed searchable forever, and an agent could still load its body through `get_skill_content`.

## What

`SkillCatalog.replaceAll(skills)` / `replace_all(skills)` — the batch becomes the entire catalog.

- **Replace means replace.** Ids absent from the batch are removed, including ones registered in-process. A host mixing local and remote skills composes the batch itself: `replaceAll([...local, ...remote])`. Ownership policy stays in the host, not the core.
- **Mutates in place.** `ratel()` closes over one `SkillCatalog` in `r.skills`, each `adaptTo` view, and every capability tool from `modelTools()`; rebinding the reference would strand all of them. `advertiseSkills: true` already pins the `search_capabilities` description, so a reload can't bust the prompt cache.
- **Two-phase, like `register`.** The corpus swap commits synchronously; the embedding pass is the awaitable. On failure the new corpus is live and BM25 ranks it, while semantic search reports `EmbeddingsNotBuilt` until a later pass succeeds. One rule to learn, not two.
- **The diff is the point.** Only removed ids' vectors are dropped and only re-worded ids invalidated; a body-only edit keeps its vector (bodies aren't embedded). Reloading an unchanged catalog costs **zero** embedding calls.
- Returns `ReplaceOutcome` (`added`/`removed`/`updated`/`unchanged`) so a source can report a reload without draining the trace stream.
- Churn telemetry falls out of the same diff — `add`/`remove` for real changes only, reusing the existing `ChurnKind::Remove`. A no-op reload emits nothing.

Tools are deliberately out of scope: a cloud-sourced tool carries no local executor, which is a separate question.

### One subtle constraint

Dropping a removed id's vector is load-bearing rather than tidiness. The dense cache's built-ness guard compares *counts*, so a stale vector left behind could offset a new, unembedded id, let the guard pass, and make a semantic search **silently omit** that skill. That is exactly the state a failed embedding pass leaves behind, so it interacts directly with the two-phase decision. Pinned by `replace_all_drops_the_vector_of_a_removed_id`.

## Layers

| | |
|---|---|
| `src/core` | `SkillRegistry::replace_all -> ReplaceOutcome`; `Skill` derives `PartialEq` |
| bindings | `replaceAll` (napi, via `write_registry` so it inherits the registry-busy guard); `_replace_all` (pyo3) + `.pyi` |
| `src/sdk/ts` | `SkillCatalog.replaceAll`, internal `replaceAllItems`, `ReplaceOutcome` exported |
| `src/sdk/python` | `replace_all` on catalog + registry facade, `ReplaceOutcome` dataclass |
| docs | ADR-0014, plus a section in both SDK READMEs |

ADR-0014 implements the hydration half of ADR-0003, which accepted a loader that "pulls a published catalog and hydrates the local registries" but left the mutation primitive unspecified. **ADR-0003 is unchanged and stays Accepted.**

## Verification

| | |
|---|---|
| `cargo test --workspace` | 171 pass |
| `cargo clippy -D warnings` / `fmt --check` / `doc` | clean |
| `pnpm -r build` / `typecheck` / `lint` / `test` | 446 pass, 1 skipped |
| `ruff` / `mypy` / `pytest` | clean; 147 pass, 2 skipped |

23 new tests. The core ones went red→green properly (9 failed to compile first). Beyond unit coverage, `ratel.test.ts` pins the end-to-end story: take `modelTools()` once, reload, then assert the `search_capabilities` description is byte-identical, a dropped skill returns the structured error through the already-exposed `get_skill_content`, a new one loads, and the dropped skill is gone from `recall()`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/adr/0015-whole-catalog-skill-reload.md
+++ b/docs/adr/0015-whole-catalog-skill-reload.md
@@ -33,9 +33,14 @@ be a small change rather than an architecture:
 - **The model-facing payload is already pinned.** `ratel()` builds `search_capabilities` with
   `advertiseSkills: true`, so the tool description does not depend on whether the catalog is
   empty. A reload cannot perturb the prompt cache.
-- **The registries are already lock-guarded.** The bindings hold the registry behind a write
-  lock and refuse a mutation while a dense operation is in flight, so a corpus swap is never
-  observed torn and two overlapping reloads cannot interleave.
+- **The registries already refuse a racing mutation.** Each binding rejects a mutation while a
+  dense operation is in flight, so a corpus swap is never observed torn and two overlapping
+  reloads cannot interleave. Where that guard lives differs by surface, and the difference
+  matters to anyone adding a mutation: the napi binding holds the registry in an
+  `Arc<RwLock<_>>` and goes through `write_registry`, which checks the pending-dense counter and
+  takes the lock; the pyo3 binding owns its registry unlocked, and the equivalent check
+  (`_raise_if_busy` against `_dense_pending`) lives in the Python facade above it, not in the
+  native layer.
 
 ## Decision
 
--- a/docs/adr/0015-whole-catalog-skill-reload.md
+++ b/docs/adr/0015-whole-catalog-skill-reload.md
@@ -1,0 +1,121 @@
+# 15. Whole-catalog skill reload: `replaceAll` mutates the catalog in place
+
+Date: 2026-07-27
+
+## Status
+
+Accepted
+
+Implements the catalog-hydration half of [ADR-0003](0003-catalog-source-interface.md), which
+accepted a source loader that "pulls a published catalog and hydrates the local registries" but
+left the mutation primitive unspecified. ADR-0003 is unchanged and remains Accepted; this ADR
+adds the missing primitive and fixes its semantics.
+
+## Context
+
+A catalog source — the managed cloud client first — reloads the full skill catalog periodically
+and hands it to the SDK. ADR-0003 already settled that sync is a whole-catalog pull rather than
+a delta protocol, so what arrives is always the complete set.
+
+The SDK had no way to apply it. `SkillCatalog` was append-only: `register` replaces an id in
+place, and nothing removes one. A reload could only ever grow the corpus, so a skill deleted
+upstream stayed searchable forever — and an agent could still load its body through
+`get_skill_content`.
+
+Replacing the `SkillCatalog` *object* is not available as a fix. `ratel()` creates it once and
+closes over it in `r.skills`, in each `adaptTo` view's `base.skills`, in every capability tool
+returned by `modelTools()`, and in `recall()`; hosts hold `r.skills` directly as the documented
+escape hatch. Rebinding the reference would strand all of them.
+
+Two properties of the existing design make in-place mutation safe, and are the reason this can
+be a small change rather than an architecture:
+
+- **The model-facing payload is already pinned.** `ratel()` builds `search_capabilities` with
+  `advertiseSkills: true`, so the tool description does not depend on whether the catalog is
+  empty. A reload cannot perturb the prompt cache.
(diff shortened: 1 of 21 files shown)


kayraucklnc:
Nice work, the core diff is tight and the two dense-cache tests are genuinely load-bearing (I mutation-tested both: dropping the removal-side `invalidate` fails only `replace_all_drops_the_vector_of_a_removed_id`, and always-invalidating on update fails only `replace_all_keeps_the_vector_when_only_the_body_changed`).

One thing I'd fix before merge, on the TS side.

### `replaceAll`'s documented "read without awaiting" path is an unhandled rejection

`replaceAll` starts `buildDense()` eagerly and hangs the counts off the resulting promise ([`skill-catalog.ts:126`](https://github.com/ratel-ai/ratel/blob/feat/skill-catalog-replace-all/src/sdk/ts/src/skill-catalog.ts#L126)). Both the `PendingReplace` doc ([L19](https://github.com/ratel-ai/ratel/blob/feat/skill-catalog-replace-all/src/sdk/ts/src/skill-catalog.ts#L19), "read them without awaiting") and the `@returns` ([L114](https://github.com/ratel-ai/ratel/blob/feat/skill-catalog-replace-all/src/sdk/ts/src/skill-catalog.ts#L114), "Read `.added`/`.removed` without awaiting") tell callers to do exactly that. If the embedding pass then fails, nothing is attached to the rejection.

Confirmed against the built SDK:

```js
const pending = catalog.replaceAll([skill("a", "alpha")]); // semantic, bad model path
console.log(pending.added, pending.removed);               // 1 0 — works as documented
// ...and then:
// !! UNHANDLED REJECTION: EmbedderError: failed to load embedding model /definitely/missing/model
```

Under Node's default `--unhandled-rejections=throw` that takes the host process down. Which is a shame, because the failure you're designing for here (embeddings fail, corpus is live, BM25 still ranks) is precisely the one a periodic source is supposed to survive and report on.

The READMEs are fine, both examples `await` inside a `try`. It's only the JSDoc that invites the bare-read path, and it's the JSDoc a host actually reads at the call site.

Python doesn't have this problem, and its docstring already says the right thing: **"Always `await` the result."** I'd just mirror that on the TS side, i.e. keep "the counts are final at call time" as the affordance, but say plainly that the returned value must still be awaited (or `.catch()`-ed) even when you only want the counts. Making the promise lazy would also work, but that's a bigger change than the problem warrants.

Two smaller things while you're in there:

- Both READMEs still say a reload "races a **dense** operation" ([ts:52](https://github.com/ratel-ai/ratel/blob/feat/skill-catalog-replace-all/src/sdk/ts/README.md#L52), [py:51](https://github.com/ratel-ai/ratel/blob/feat/skill-catalog-replace-all/src/sdk/python/README.md#L51)). `8cb4d92` widened that claim in ADR-0015 and in both docstrings to include an in-flight BM25 `search_async`, so this is that same fix, two files short.
- `PendingReplace` in Python is a frozen dataclass subclassing `ReplaceOutcome`, so `catalog.replace_all(b) == ReplaceOutcome(added=1, ...)` is `False` even when the counts match, while the awaited value compares `True`. `field(compare=False, repr=False)` on `_driver` fixes that and stops `repr()` leaking the coroutine.

Everything else ran green for me: `cargo test --workspace`, clippy/fmt, `pnpm -r build/typecheck/lint/test`, and `ruff`/`pytest`. ADR-0015 in particular is a good record, the per-binding guard-location asymmetry is exactly the kind of detail that normally gets lost.

rstagi (approved):
lgtm! There's just one test failing, can you have a look at that? After that we can merge it

PR #143, opened 2026-08-06
feat: configurable adaptive ranking

bercaakbayir:
Adds the **seed-first** path for adaptive usage ranking (ADR-0014): record what an
agent invokes while Ratel serves nothing, build an intent graph from that log
offline, inspect it, and only then switch ranking on.

Adaptive ranking learns from what the agent invokes. Once Ratel is ranking, what
the agent invokes is partly Ratel's own doing — ADR-0014 concedes the loop:
*"boosting used capabilities makes them more used."* And a fresh deployment gets no
boost at all until enough pairs accumulate. Capturing first fixes both, and gives
you enough evidence to be useful on day one.

## What's new

**Capture** — `experimentalBaselineTurn(query).invoked(id).record()` records a turn
observed while Ratel is not in the retrieval path. Buffered: nothing reaches the log
until `record()`, so a turn that fails your own quality gate is simply never written.

**Build** — `experimentalBuildIntentGraph(jsonl, options?)` turns a trace log into a
graph. Every distinct query is embedded up front so clusters form at the dense tier,
exactly as the live path would grow them. The returned graph is **detached** —
building never switches ranking on.

**Policy** — `origins` (`any` | `agent` | `baseline`) and `provenance` (`live` |
`seeded`) apply to both the live and offline paths, so what counts as evidence does
not depend on which produced the graph. Building defaults to `baseline`/`seeded`,
since that is what an offline build is; enabling live learning keeps `any`/`live`.
Values are a closed set — a typo is a compile error, not a silently different graph.

**Provenance** — `seeded_support` on each cluster records how many observations came
from a capture. After the flip, `support` grows while `seeded_support` stays put, so
the gap tells you how much still rests on the baseline.

## Distributed hosts

The single-process API assumes one process owns the turn and that the SDK owns where
the trace lands. Neither holds for a process-per-request server across many
instances, which is the first real deployment this is aimed at. Two additive seams:

- **`FnSink`** (core) / **`{ kind: "callback", sessionId, onEvent }`** (SDK) — hands
  each envelope to a closure as a JSON line, byte-identical to what the `jsonl` sink
  writes. Lines collected from many processes can be joined with newlines and passed
  straight back to `experimentalBuildIntentGraph`.
- **`experimentalRecordBaselineTurn({ query, invoked, invokedSkills })`** — records a
  turn the host reassembled from its own storage. Not just ergonomics: the chained
  builder allows an `await` between `invoked()` calls, so concurrent turns can
  interleave and break the adjacency the graph pairs on, and one turn per invocation
  counts the query N times — inflating the support that gates the flip.

Nothing shipped was modified. The existing builder is untouched, so both demos still
exercise it as a regression check, and a parity test holds the two paths together.
Defaults do the work: recording stamps `baseline`, building defaults to
`origins: baseline` / `provenance: seeded`, so a host passes no options.

Two bugs surfaced during implementation and are fixed here: the threadsafe function
kept the event loop alive, so a script that recorded a few turns never exited, and
callback delivery is asynchronous rather than inline, which a per-request host must
flush before its handler resolves.

`docs/baseline-capture-distributed.md` is the written version.

## Python

Python gains the same whole-turn API,
`experimental_record_baseline_turn(query, invoked=..., invoked_skills=...)`, with
tests mirroring the TypeScript ones: parity with the chained builder, skill coverage,
and a cross-request reassembly that builds the same graph.

It deliberately does not gain the `callback` sink. Calling into Python from the sink
means taking the GIL on the hot path, which is the blocking ADR-0007 rules out, and a
non-blocking version needs a queue and a drain thread. A per-request Python host gets
the same result from the `memory` sink and `drain_trace_events()` per request, which
is what the new test does.

## Demos

`examples/configurable-adaptive-ranking-{ts,python}` — no model or API key, pure
BM25. Ten turns across three intents, scored after each so you can watch clusters
reach full strength at different points, then the flip:

    query: "why is the build broken"
      cold (BM25 only) : docker_build > gh_run_list
      ...
    C. after seeding : gh_run_list > docker_build

The TS demo adds a fifth phase replaying the same ten turns as a per-request server
would — query stashed at search, replayed at invoke, one sink per turn — landing the
same graph. The Python demo covers phases A–D, since phase E is written around the
callback sink; its README says so and points at the drain equivalent.

## Breaking

`Origin` is now `#[non_exhaustive]` — exhaustive matches need a `_` arm. Rust core
only; no SDK surface changes.

`Confirmation` was removed from `ObservationPolicy`. It selected whether
`invoke_start` or `invoke_end` closed an observation, framed as filtering wrong tool
choices — but a trace records which tool was called, never whether calling it was
right. The choice is the only signal, so there was nothing to configure.

## Caveats worth reading before adopting

- **Every invocation is evidence, and the graph assumes it is good evidence.** Nothing
  in a trace says whether a turn went well, so nothing is filtered. Seed from an agent
  you already trust. Edge weights set only *order*, so a single wrong invocation of a
  tool the base ranker already favours is close to free and more good data does not
  dislodge it.
- **Gate on coverage**, not cluster count — held-out queries that match a cluster. The
  other numbers rise whether or not the graph generalises.
- **Callback delivery is asynchronous and lossy**, per ADR-0007. Flush before a process
  exits; treat a capture window as best-effort.
- **`SUPPORT_FULL` is not exposed.** The demos hardcode `3`. A first-class readiness
  API is still to come.
- **Rebuilding is O(whole log).** Fine nightly, wasteful per turn.

## Release

Ships as `@ratel-ai/sdk@0.6.1-rc.2` (published, all five platform binaries). Core is
unreleased and rides inside the TS binary; it can be cut on its own `core-v*` tag
later. Python stays at `0.6.0` here and needs its own `sdk-py-v*` tag whenever you
want to ship it.

## Verification

- Rust: 354 tests, clippy `-D warnings`, fmt — all clean
- TS: 311 tests, typecheck, type-tests, biome, typedoc — all clean
- Python: 202 tests, ruff, mypy — all clean
- Protocol conformance: 40/40, unchanged (`seeded_support` is omitted when zero, so a
  live-only graph serializes byte-identically to one produced before the field existed)
- Both demos run end to end
- Published tarball verified from a clean install: callback sink, per-turn session ids,
  `baseline` origin, graph build, ranking active

## Docs

- `docs/adr/0007-telemetry-two-streams.md` — amended for the host-owned destination
- `docs/baseline-capture-distributed.md` — new guide for multi-instance hosts

## After review

- Phase E claimed parity with phase B but read the graph after live learning had
  already added an observation to it. It reported success for the wrong reason.
  Now snapshots the seeded result before that point.
- `invokedSkills` had no test coverage. Added one, verified by deleting the
  implementation and confirming the test fails.
- Phase E's pending-query map was set, read and deleted inside one loop
  iteration, so it demonstrated no cross-request handoff. Now driven by a request
  stream where two searches land before either invocation, and a turn whose
  query is gone is dropped rather than throwing.
- Four copies of the learner-wrapping match, one per sink-install path, collapsed
  into a single helper. The generated `index.d.cts` is byte-identical across the
  refactor, and inverting the helper fails the guard test.
- Removed an unused accessor added by this PR.
- Corrected a documentation bug the review surfaced: I had written in three places
  that turns sharing a session id would cross-pair. Testing it showed the opposite,
  since recording a turn whole emits its search and invoke adjacently and replay
  tracks one pending query per session in log order. A per-turn id protects you only
  once lines from several producers are merged without preserving that adjacency.
  Fixed in the demo, the guide and the shipped JSDoc, and pinned by a test.


kayraucklnc (approved):
Reviewed the full branch, resolved the conflicts with main's runtime-events lane, and verified the merged tree end to end: Rust 490 + clippy + fmt, TS 800 + typecheck + biome + typedoc, Python 316 + 97 + ruff + mypy, protocol conformance 40/40, both demos run and produce the flip. CI green on all legs including both verify platform legs.

Three defects fixed on top: the Python demo could not run on the Python its pyproject declares (nested f-string needing PEP 701), the callback sink x runtime-events interaction was untested (mutation-verified regression test added), and a CHANGELOG documented a keyword that had been removed. ADR-0014 amended to record the seed-first path it never mentioned.

PR #144, opened 2026-08-07
perf(core): cache BM25 index across searches, rebuild only on catalog mutation

rstagi:
## Problem

User feedback: BM25 + embeddings indexing causes CPU spikes even with a remote embeddings endpoint. Root cause on the BM25 side: every search rebuilt the whole engine — `bm25_search` tokenized + stemmed the full corpus twice per call (bm25 crate fit pass + embed pass), and `bm25_docs()` re-derived `searchable_text` (JSON-schema walk + clones) for every item on every search. Fires 2-6x per agent turn (capabilities search = tools + skills buckets; hybrid pays the BM25 arm on top of dense).

## Fix

Index once per **catalog state**, not per search:

- `Bm25Index`: build/query split of the old one-shot path. Build = full-corpus tokenize + index; query = tokenize query + score candidates.
- `Bm25Cache`: `Mutex<Option<Arc<Bm25Index>>>` dirty-flag holder (DenseCache precedent). First search after a mutation rebuilds via the same full-corpus code path; later searches reuse. Searches rank outside the lock.
- Invalidated by the three corpus mutators: `ToolRegistry::register`, `SkillRegistry::register`, `SkillRegistry::replace_all`. `replace_all` keeps the index when no *indexed* text changed (unchanged or body-only reloads) — same diff the dense cache keys on.
- Hybrid arms reuse the cached index (previously inlined a second `searchable_text` derivation).
- Deliberately no incremental `upsert`: the crate freezes `avgdl` at fit time, upserted scores drift vs a fresh build. Full rebuild-on-dirty keeps scores byte-for-byte (ADR-0011).

## Numbers

1,000 tools, 50 searches, release build, same machine:

| | first search | each warm search | 50 searches total |
|---|---|---|---|
| main | 24.8ms | 25.4ms | 1.27s |
| this PR | 20.4ms (build) | 262µs | 13ms |

~97x on warm searches; the one-time build cost is unchanged.

## Determinism

Scores stay byte-identical: every rebuild is a full-corpus build with the same k1/b, same rank-all-then-truncate, same `(score desc, id asc)` tie-break; corpus stats (avgdl/IDF) don't depend on iteration order. Pinned by `warmed_registry_hits_are_byte_identical_to_a_fresh_registry` (ids AND f32 scores) and the existing tie-break tests. ADR-0011/0014 amended in place for the `bm25_search` → `Bm25Index::search` rename.

## Tests

TDD throughout (red → green per cycle). Mutation-verified: reverting the search paths to build-per-call fails `bm25_cache_is_warmed_by_search_and_dropped_by_register` and the skill-side twin — the tests warm via public `search()` and assert the builder never re-runs, so the wiring can't silently regress. Staleness guards: register/replace_all after warmed searches must be visible in the immediately following search; unchanged/body-only `replace_all` must keep the cached index.

`cargo test --workspace`, `clippy --all-targets -D warnings`, `fmt --check` all green.

### DIFF
--- a/docs/adr/0011-selectable-retrieval-methods.md
+++ b/docs/adr/0011-selectable-retrieval-methods.md
@@ -28,10 +28,11 @@ is chosen per catalog (a construction-time default) or per call (an explicit ove
 override wins. Across the SDKs the identifier is the string `"bm25" | "semantic" | "hybrid"`,
 parallel to `SearchOrigin`.
 
-- **BM25** is unchanged from ADR-0004 (`bm25_search`, `k1 = 0.9`, `b = 0.4`) and stays the
-  default. The legacy `search` / `search_with_origin` entry points keep their infallible
+- **BM25** is unchanged from ADR-0004 (`Bm25Index::search`, `k1 = 0.9`, `b = 0.4`) and stays
+  the default. The legacy `search` / `search_with_origin` entry points keep their infallible
   `Vec<SearchHit>` signature and BM25 behavior byte-for-byte — upgrading callers need no code
-  change.
+  change. The index is cached per registry (`Bm25Cache`) and rebuilt in full on the first
+  search after a corpus mutation, so scores always equal a fresh build's.
 - **Semantic** embeds the query with a local `BAAI/bge-small-en-v1.5` model (pure-Rust Candle,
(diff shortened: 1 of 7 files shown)


faustoq (approved):
LGTM. Big improvement! hybrid's duplicate `searchable_text` derivation gets fixed as a side effect too.

PR #147, opened 2026-08-10
feat(core): cache embeddings

faustoq:
## Problem

Semantic / hybrid catalogs currently rebuild document embeddings at runtime.

For static or mostly-static Tool / Skill corpora, every cold start pays that document-inference cost again: local CPU time or remote endpoint latency/cost.

## Solution

Add immutable **build-time embedding artifacts** (`RAT1`) that warm the runtime dense cache without re-embedding unchanged catalog documents.

Build once:

```ts
await experimentalBuildEmbeddingArtifact({
  output: "/opt/ratel/catalog.ratel-embeddings",
  embedding,
  tools,
  skills,
});
```

Reuse at runtime:

```ts
const catalog = new ToolCatalog({
  method: "hybrid",
  embedding,
  experimentalEmbeddingArtifact: {
    path: "/opt/ratel/catalog.ratel-embeddings",
    onMiss: "error",
  },
});
```

`onMiss` controls documents that cannot be reused:

`"error"`: (default) fail with the exact missing ids
`"embed"`: reuse hits and embed only missing / changed documents

Artifacts are immutable build output; there is no runtime write-through.

Artifact persistence is host-owned: the core artifact APIs only produce and consume bytes and perform no artifact filesystem I/O.

## RAT1

RAT1 v1 stores only the data needed to safely reuse embeddings:

- Tool / Skill kind + stable id
- searchable-projection hash
- normalized vector
- projection version
- model fingerprint + dimension

It does not contain descriptions, skill bodies, executors, credentials or model weights.

One artifact may contain both Tools and Skills; identity is `(kind, id)`. Registries ignore entries for the other kind and artifact-only entries.

Tool-only, Skill-only, mixed and empty artifacts are supported. Merge rejects model / dimension / projection mismatches and duplicate `(kind, id)` entries.

## Numbers

Fresh-process release benchmark using Ratel's default local embedding model:

- deterministic corpus: **100 tools + 20 skills**
- same machine and embedding configuration
- Python native extension built with `maturin develop --release`
- a new Python process for every measured sample
- **20 measured samples per variant**
- 2 warm-up pairs discarded
- normal/RAT1 order alternated between pairs, with a 5 s settling interval between fresh worker processes
- prebuilt RAT1 reused across measured runs; artifact build excluded

|                                         | normal registration | RAT1 warm        |
| ----------------------- | ------------------: | ---------------: |
| documents                     |                           120 |                     120 |
| median registration       |                       9.04 s |           305.2 ms |
| p10–p90 registration    |              8.77–9.55 s | 44.2–366.7 ms |
| median process wall     |                        9.19 s |           457.2 ms |
| p10–p90 process wall    |             8.89–9.73 s | 172.3–568.1 ms |
| registration speedup    |                               — |                 29.6x |
| process-wall speedup    |                             — |                 20.1x |
| artifact size                    |                              — |         190,064 B |

The RAT1 artifact took 9.72 s to build once; that build cost is excluded from the registration and process-wall measurements above.

With full artifact coverage, normal registration infers all 120 corpus documents. RAT1 reuses their vectors and performs no corpus-document inference.
For the Local model used here, both paths still pay model initialization/load under the current runtime semantics; RAT1 removes corpus-document inference, not model load.

## Concurrency

Artifact build does not take the dense cache's operation lock, so existing semantic search may continue.

Artifact warm mutates the dense cache and is serialized. With `onMiss: "embed"`, reuse + missing-document inference stay under one dense operation lock.

## SDKs

TypeScript and Python expose the same build / warm / `onMiss` contract, including path or in-memory bytes and typed artifact failures.

## Errors

Artifacts fail closed on malformed/checksum-invalid RAT1, unsupported versions, model/dimension mismatch and incompatible merges.

`onMiss` option controls documents that cannot be reused:

- `"error"` (default): fail with the exact missing ids
- `"embed"`: reuse hits and embed only missing / changed documents

Build also rejects an embedder returning a different number of vectors than inputs.

## Validation

* Rust core tests + clippy
* TypeScript build / typecheck / lint / full test suite
* Python pytest / ruff / mypy
* native bindings clippy

Architecture is documented in [ADR-0018](https://github.com/ratel-ai/ratel/blob/feat/embed-artifact-caching/docs/adr/0018-build-time-embedding-artifacts.md)


src/core/src/embedding_config.rs
```diff
@@ -395,33 +556,33 @@ impl EmbeddingModel {
...
     /// stamped on the dense cache after load.
-    pub(crate) fn configured_fingerprint(&self) -> String {
+    pub(crate) fn configured_fingerprint(&self) -> Result<String, EmbedderError> {
         let base = match self {
             EmbeddingModel::Default => huggingface_fingerprint(DEFAULT_REPO, DEFAULT_REVISION),
             EmbeddingModel::HuggingFace { repo, revision, .. } => {
                 huggingface_fingerprint(repo, revision.as_deref().unwrap_or("main"))
             }
-            EmbeddingModel::Local { path, .. } => local_fingerprint(&path.display().to_string()),
+            EmbeddingModel::Local { path, .. } => local_fingerprint(&local_model_content_id(path)?),
```
kayraucklnc: This hashes the **entire weights file** on every process start. I measured ~795ms per 400MB in release (~500 MB/s), so ~2s for a 1GB local model.

The memo is process-local, so it never helps a cold start, which is exactly the case this PR exists to improve. And every `Local` model user pays it whether or not they ever touch an artifact.

Suggestion: only compute the content id when an artifact is actually configured, or hash config + tokenizer plus weights `(len, mtime)` and do the full digest on demand.

src/core/CHANGELOG.md
```diff
@@ -6,6 +6,14 @@ The format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) and
 
 ## [Unreleased]
 
+### Added
+
+- **Build-time embedding artifacts (ADR-0017).** Binary RAT1 format (magic/version/length/checksum, projection header, Tool/Skill entries with id + projection hash + L2-normalized vector). `ToolRegistry` / `SkillRegistry` build a single-kind artifact and warm the dense cache from bytes (`OnArtifactMiss::Error` or `Embed`). `merge_embedding_artifacts` combines compatible parts into one mixed RAT1. Public exports: `ArtifactEntryKind`, `ArtifactError`, `merge_embedding_artifacts`, `OnArtifactMiss`, `ArtifactWarmError`, `WarmError`. Core stays filesystem/I/O-free on this surface — hosts own persistence.
+
+### Changed
+
+- **Local embedding model fingerprints are content-derived.** `EmbeddingModel::Local` identity is a SHA-256 of the files Candle loads, not the directory path, so a build-time embedding artifact can still validate the model after a Docker remount. HuggingFace and endpoint fingerprints are unchanged. **Note:** an `IntentGraph` persisted under a previous path-based Local fingerprint will report `ModelMismatch` until centroids are rebuilt under the active model.
```
kayraucklnc: This is a **breaking change to a stable surface**, not just "Changed". Any persisted `IntentGraph` built under a `Local` model reports `ModelMismatch` after upgrading.

Given the additive / non-breaking evolution rule in AGENTS.md, this needs an explicit breaking call-out and a minor bump, not a patch. Good catch documenting it at all, it just needs to be louder.

src/core/src/dense_cache.rs
```diff
@@ -164,6 +224,147 @@ impl DenseCache {
...
+        }
+
+        if reused.is_empty() {
+            return Ok(WarmOutcome {
+                reused: Vec::new(),
+                missing,
+            });
+        }
+
+        let embedder = self.resolve_embedder(sink)?;
```
kayraucklnc, line 307: Warm still loads the embedding model here. Skipping document inference is real, but for a local model the model load is most of what's left, so the artifact saves less than the PR description suggests.

Not a bug (you need the model for query embedding anyway), but it's the mechanical reason the benchmark speedup doesn't reproduce on a release build.

src/core/src/embedding_artifact.rs
```diff
@@ -0,0 +1,1297 @@
...
+
+    let dim = vectors.first().map(Vec::len).unwrap_or(0);
+    for ((id, _), vector) in rows.iter().zip(&vectors) {
+        if vector.len() != dim {
+            return Err(ArtifactError::InconsistentVectorWidth {
+                expected: dim,
+                got: vector.len(),
+            });
+        }
+        if dim != 0 && !is_unit_normalized(vector) {
```
kayraucklnc: Edge case: if the embedder returns empty vectors, `dim` is 0, this check is skipped, and you get a structurally valid artifact full of zero-length vectors.

Warm accepts it (`validate_batch` derives dim from the first vector) and `dense_search::cosine` zips to the shorter slice, so **every score silently becomes 0.0** with no error anywhere.

Rejecting `dim == 0` when `entries` is non-empty would match how defensive the rest of this file is.

src/sdk/ts/CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ The format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) and
 
 ## [Unreleased]
 
+### Added
+
+- **Experimental build-time embedding artifacts (ADR-0017).** `experimentalBuildEmbeddingArtifact` builds a mixed Tool+Skill RAT1 file (hosts own persistence. Core stays I/O-free on bytes). Catalogs accept `experimentalEmbeddingArtifact: { path } | { bytes }` (default `onMiss: "error"`) and warm it on `register` / `replaceAll` for any search method. Before eager document embedding on semantic/hybrid. `ToolRegistry` / `SkillRegistry` expose `buildEmbeddingArtifact` and `warmEmbeddingsFromArtifact`. `mergeEmbeddingArtifacts` combines compatible single-kind parts. Failures are typed `ArtifactWarmError` (stable `.code`, e.g. `"Incomplete"` with `.missing`). Shipped behind an `experimental` prefix (the API may change until it graduates).
```
kayraucklnc: `mergeEmbeddingArtifacts` isn't exported from the SDK. It's only imported internally from `../native/index.cjs` in `embedding-artifact.ts:2`, and it's absent from `dist/index.d.ts`.

Same on the Python side: the tests import `merge_embedding_artifacts` from `ratel_ai._native`, yet `ratel_ai` publicly exports `IncompatibleMergeError`, an exception no public Python function can raise.

Either export merge on both SDKs, or drop it from both changelogs and unexport `IncompatibleMergeError`.

kayraucklnc (commented):
Reviewed this end to end: checked it out into a worktree and actually ran it. Everything is green (cargo test 295 passed, clippy + fmt clean, pnpm build/typecheck/lint/test clean, pytest 171 passed). I also mutation-tested the two safety gates (projection-hash match, model-fingerprint gate) and both turn tests red when broken, so the tests are real. Nice work on the format itself.

Three things to fix before this ships.

### 1. The benchmark numbers are from a debug build

`maturin develop` installs a **debug** extension, and Candle is ~35x slower there. Same SDK, same 120-tool corpus, same machine, warm HF cache:

| build | register + embed 120 docs |
|---|---|
| `maturin develop` (debug) | **46.35s** |
| `maturin develop --release` | **1.32s** |
| Rust native, `cargo test --release` | **1.37s** |

So the "~7 min" baseline and the 694x / 513x speedups don't reproduce. Against a release build the local-model speedup is about **2x**, and most of what remains is model load, which the warm path also pays (see my comment on `dense_cache.rs`).

The feature is still worth shipping, but the real win is **endpoint-backed embeddings**, where you skip actual network round-trips and per-token cost. The table doesn't measure that case at all. Please re-measure on release before these numbers go anywhere near a changelog or a blog post.

### 2. The local fingerprint change is breaking, and it costs cold-start time

Two separate problems, both inline below on `embedding_config.rs` and `CHANGELOG.md`.

### 3. The changelogs and ADR-0017 describe APIs that aren't exported

`mergeEmbeddingArtifacts` / `merge_embedding_artifacts` are internal on both SDKs (inline comment below). And ADR-0017 says both SDKs expose typed artifact failures, but TypeScript has no `ArtifactError` at all: `experimentalBuildEmbeddingArtifact` routes build failures through `mapEmbedderError`, so a corrupt input surfaces as a plain `Error`. Python has `ArtifactError` and `IncompatibleMergeError`. Either add them to TS or soften the claim.

---

The rest of my notes are inline and are nice-to-haves. The format design is genuinely solid: checksum + length-prefixed fields, checked-math pre-allocation guards, trailing-byte rejection at both levels, and the test proving descriptions/bodies/credentials never reach the bytes.

src/sdk/python/native/src/lib.rs
```diff
@@ -49,6 +70,65 @@ fn map_embedder_err(e: core::EmbedderError) -> PyErr {
...
+        }
+        err
+    })
+}
+
+fn map_artifact_warm_err(e: CoreArtifactWarmError) -> PyErr {
+    match e {
+        CoreArtifactWarmError::Incomplete { missing } => {
+            let message = format!(
+                "embedding artifact incomplete for the current corpus: {} id(s) missing ({})",
```
kayraucklnc: This re-formats by hand the exact message `artifact_warm.rs:87` already produces, while the TS binding uses `error.to_string()`. If the core `Display` ever changes, Python and TypeScript silently drift apart.

Capture `e.to_string()` before destructuring `missing` and pass it through.

src/sdk/ts/src/ratel.ts
```diff
@@ -18,21 +18,26 @@ import {
...
    * catalogs — see {@link ToolCatalogOptions.embedding}. A string is a local model
    * directory path; every other source is a keyed object (`{ huggingface }`,
    * `{ ollama }`, `{ url, model, apiKeyEnv }`). Omit to use the built-in default
-   * model. `await r.tools.register(...)` awaits the embedding pass and rejects if
-   * it fails, so errors surface at registration. */
+   * model. `await r.tools.register(...)` awaits dense preparation (artifact warm
+   * or embedding) and rejects if it fails, so errors surface at registration. */
   embedding?: EmbeddingSpec;
+  /** Build-time embedding artifact forwarded to both catalogs (default `onMiss: "error"`). */
+  experimentalEmbeddingArtifact?: ExperimentalEmbeddingArtifact;
```
kayraucklnc, line 53: Worth saying here that the artifact has to cover **both** tools and skills, since this forwards one artifact to both catalogs.

I hit this: with the default `onMiss: "error"`, a tools-only artifact makes `skills.register()` reject with `ArtifactWarmError: ... 1 id(s) missing (frontend-slides)`. Failing closed is correct, but one sentence in this doc comment prevents the obvious first-use failure.

src/sdk/ts/src/embedding-artifact.ts
```diff
@@ -0,0 +1,126 @@
...
+    throw new TypeError(
+      `experimentalEmbeddingArtifact has unknown keys: ${unknown.sort().join(", ")}`,
+    );
+  }
+  const hasPath = Object.hasOwn(record, "path");
+  const hasBytes = Object.hasOwn(record, "bytes");
+  if (hasPath === hasBytes) {
+    throw new TypeError("experimentalEmbeddingArtifact requires exactly one of 'path' or 'bytes'");
+  }
+  const onMissRaw = Object.hasOwn(record, "onMiss") ? record.onMiss : "error";
```
rstagi: - [x] Treat explicit `undefined` as absent (`record.onMiss !== undefined`, same for `bytes`), not key presence via `Object.hasOwn`

`{ path, onMiss: cond ? "embed" : undefined }` and `{ path, bytes: undefined }` typecheck (no `exactOptionalPropertyTypes`) but throw here on every `register()`, so the dense cache is never prepared.

src/core/src/embedding_artifact.rs
```diff
@@ -0,0 +1,1297 @@
...
+    }
+
+    let mut entries = Vec::with_capacity(entry_count);
+    for _ in 0..entry_count {
+        let kind = read_kind(payload, &mut cursor)?;
+        let id = read_utf8(payload, &mut cursor)?;
+        let projection_hash = read_fixed::<32>(payload, &mut cursor)?;
+        let mut vector = Vec::with_capacity(dim);
+        for _ in 0..dim {
+            vector.push(read_f32(payload, &mut cursor)?);
```
rstagi: - [x] Validate decoded vectors (finite + unit-norm) at load, or state the trust model explicitly in ADR-0017

Build enforces `is_unit_normalized`, but load/warm commits decoded vectors verbatim: a checksum-valid artifact with NaN/Inf or `[3.0, 0.0]` warms fine and silently corrupts cosine ranking (dense_search assumes unit vectors). SDKs accept raw `bytes`, so third-party writers are reachable. Related: warm seeds `state.dim` from the self-declared header dim.

src/sdk/ts/src/registry.ts
```diff
@@ -96,6 +96,35 @@ export class ToolRegistry {
...
   }
 
+  /**
+   * Build a binary embedding artifact from the registered corpus (ADR-0017).
+   * Embedding for the artifact is independent of the mutable dense cache —
+   * it does not consume or update cached vectors. Returns a Node `Buffer`
+   * suitable for `fs.writeFile`. Independent of the registry's search method
+   * (works with `"bm25"` as well as `"semantic"`/`"hybrid"`).
+   */
+  async buildEmbeddingArtifact(): Promise<Buffer> {
```
rstagi: - [x] Decide experimental naming for `buildEmbeddingArtifact` / `warmEmbeddingsFromArtifact` before an RC ships (same for Python `build_embedding_artifact` / `warm_embeddings_from_artifact`)

`ToolRegistry` / `SkillRegistry` are public exports in both SDKs, so these land un-prefixed while the rest of the feature is `experimental*`. Renaming after release is a break; the CHANGELOG claim "shipped behind an experimental prefix" is inaccurate for these.

rstagi (commented):
Ran the branch in an isolated worktree: cargo build/test/clippy, pnpm build/typecheck/lint/test, maturin + ruff/mypy/pytest all green. Also exercised end-to-end against a counting embedding endpoint: build makes one batched embed call, warm makes zero doc-embed calls on full coverage, `onMiss: "error"` raises typed `Incomplete` with exact ids, `onMiss: "embed"` embeds only the missing/changed doc, corrupt bytes fail closed. No blockers; inline comments below cover recommended fixes plus cleanup nits.

src/sdk/ts/CHANGELOG.md
rstagi: - [x] `mergeEmbeddingArtifacts` is not exported from the package entry: either export it or drop it from the entry
- [x] Fix the sentence fragment "Before eager document embedding on semantic/hybrid."

src/sdk/python/CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ The format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) and
 
 ## [Unreleased]
 
+### Added
+
+- **Experimental build-time embedding artifacts (ADR-0017).** `experimental_build_embedding_artifact` builds a mixed Tool+Skill RAT1 file (hosts own persistence. Core stays I/O-free on bytes). Catalogs and registries accept `experimental_embedding_artifact` (`path` or `bytes`; default `on_miss="error"`) and warm it on `register` / `replace_all` for any search method. Before eager document embedding on semantic/hybrid. `build_embedding_artifact` and `warm_embeddings_from_artifact` are available on the registries; merge of compatible parts raises `IncompatibleMergeError` when fingerprints/dims disagree. Failures are typed `ArtifactWarmError` / `ArtifactError`. Shipped behind an `experimental_` prefix. The API may change until it graduates.
```
rstagi: - [x] Merge / `IncompatibleMergeError` are not part of the public `ratel_ai` surface (`merge_embedding_artifacts` lives in `_native` only): export or reword
- [x] Same sentence fragment as the TS entry ("Before eager document embedding on semantic/hybrid.")

src/core/CHANGELOG.md
```diff
@@ -6,6 +6,14 @@ The format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) and
 
 ## [Unreleased]
 
+### Added
+
+- **Build-time embedding artifacts (ADR-0017).** Binary RAT1 format (magic/version/length/checksum, projection header, Tool/Skill entries with id + projection hash + L2-normalized vector). `ToolRegistry` / `SkillRegistry` build a single-kind artifact and warm the dense cache from bytes (`OnArtifactMiss::Error` or `Embed`). `merge_embedding_artifacts` combines compatible parts into one mixed RAT1. Public exports: `ArtifactEntryKind`, `ArtifactError`, `merge_embedding_artifacts`, `OnArtifactMiss`, `ArtifactWarmError`, `WarmError`. Core stays filesystem/I/O-free on this surface — hosts own persistence.
```
rstagi: - [x] Add `ParseOnArtifactMissError` to the public-exports list; this PR adds it to the public API too.

src/core/src/dense_cache.rs
```diff
@@ -164,6 +224,147 @@ impl DenseCache {
...
+        let existing_dim = self.state.lock().expect("dense cache mutex poisoned").dim;
+        let expected_dim = validate_batch(
+            &staged,
+            reused.len(),
+            Some(existing_dim.unwrap_or(header.dim)),
+            &header.model_fingerprint,
+        )?;
+
+        let mut state = self.state.lock().expect("dense cache mutex poisoned");
+        if let Some(built) = &state.built_fingerprint
```
rstagi, line 339: - [x] Add a test warming an already-built cache

The existing-dim guard and this second `built_fingerprint` mismatch branch are never exercised by any test.

src/sdk/ts/src/embedding-artifact.test.ts
```diff
@@ -0,0 +1,455 @@
...
+      experimentalBuildEmbeddingArtifact({
+        output: join(tmpdir(), "ratel-missing-parent", "nope", "out.ratel-embeddings"),
+        tools: [],
+        skills: [],
+      }),
+    ).rejects.toMatchObject({ code: "ENOENT" });
+  });
+});
+
+describe("catalog experimentalEmbeddingArtifact", () => {
```
rstagi, line 264: - [x] Add a test for `register()` called twice on a catalog with `experimentalEmbeddingArtifact`

The option's docstring documents the re-resolve/re-warm-per-register contract, but no test pins it (both SDKs).

src/core/src/tool_registry.rs
```diff
@@ -1840,4 +1908,354 @@ mod tests {
             .collect();
         assert_eq!(stages, vec!["bm25", "dense", "rrf"]);
     }
+
+    /// Fixed-identity stub used to serialize artifacts for warm tests.
+    struct ArtifactBuildStub {
```
rstagi: - [x] Share the artifact test stubs instead of copy-pasting

`ArtifactBuildStub`, `PanicOnEmbedStub`, the counting/failing embedders, and the `unit*` helpers are duplicated near-verbatim across dense_cache.rs, tool_registry.rs, and skill_registry.rs (~200 lines).

src/core/src/embedding.rs
```diff
@@ -657,30 +658,15 @@ impl CandleEmbedder {
...
-            if !p.exists() {
-                return Err(EmbedderError::Load {
-                    model: name.clone(),
-                    source: format!(
-                        "missing {f} in {name} — a fast tokenizer.json is required; run \
-                         tokenizer.save_pretrained() upstream, or serve the model via an endpoint"
-                    ),
-                });
+        let files = resolve_local_identity_files(dir)?;
+        let (config_path, tokenizer_path, weights_path) = match files.as_slice() {
```
rstagi: - [x] Return a typed struct (config / tokenizer / weights / optional pooling) from `resolve_local_identity_files` instead of a positional `Vec<PathBuf>`

The positional contract forces this slice pattern + `unreachable!` on the consumer.

src/core/src/embedding_artifact.rs
```diff
@@ -0,0 +1,1297 @@
...
+            format_version,
+            projection_version,
+            model_fingerprint,
+            dim,
+        },
+        entries,
+    ))
+}
+
+fn write_kind(out: &mut Vec<u8>, kind: ArtifactEntryKind) -> Result<(), ArtifactError> {
```
rstagi: - [x] Drop the dead `Result` plumbing: `write_kind` (and `assemble_file`) can never fail.

src/sdk/python/native/src/lib.rs
```diff
@@ -49,6 +70,65 @@ fn map_embedder_err(e: core::EmbedderError) -> PyErr {
...
+        other => ArtifactError::new_err(other.to_string()),
+    }
+}
+
+fn artifact_warm_pyerr(code: &str, message: String, missing: Option<Vec<String>>) -> PyErr {
+    Python::with_gil(|py| {
+        let err = ArtifactWarmError::new_err(message);
+        let value = err.value(py);
+        let _ = value.setattr("code", code);
+        let _ = value.setattr("name", "ArtifactWarmError");
```
rstagi: - [x] Drop the JS-style `name` attribute on the Python `ArtifactWarmError`

`type(e).__name__` already covers it; this is a TS-ism ported into the Python surface.

src/sdk/python/native/src/lib.rs
rstagi: - [x] Use the core error's own `Display` for `Incomplete` instead of rebuilding the string here

Hand-duplicated format will drift from `ArtifactWarmError::Incomplete`'s message.

src/sdk/python/ratel_ai/embedding_artifact.py
```diff
@@ -0,0 +1,124 @@
...
+) -> tuple[bytes, OnArtifactMiss]:
+    """Load artifact bytes and resolve ``on_miss`` (default ``"error"``)."""
+    path, raw, on_miss = _validate_embedding_artifact(config)
+    if raw is not None:
+        return raw, on_miss
+    assert path is not None
+    return await asyncio.to_thread(Path(path).read_bytes), on_miss
+
+
+async def experimental_build_embedding_artifact(
```
rstagi: - [x] Move `experimental_build_embedding_artifact` above the private helpers

Module-namesake function first, helpers after first reference (repo file-ordering convention).

src/core/src/tool_registry.rs
```diff
@@ -1840,4 +1908,354 @@ mod tests {
...
+    #[test]
+    fn build_embedding_artifact_round_trips_via_warm() {
+        let a = tool("read_file", "read a file");
+        let b = tool("delete_file", "delete a file");
+        let builder = Arc::new(FpCountingEmbedder::new("fp-warm"));
+        let mut reg_a = with_embedder(builder.clone());
+        reg_a.register(tool("read_file", "read a file"));
+        reg_a.register(tool("delete_file", "delete a file"));
+        let bytes = reg_a.build_embedding_artifact().unwrap();
+        assert!(builder.docs() >= 2, "build embeds the corpus once");
```
rstagi: - [x] Tighten to `assert_eq!(builder.docs(), 2)`

`>= 2` also passes if build embeds every document twice, contradicting the assert message "build embeds the corpus once".

src/core/src/dense_cache.rs
faustoq: Addressed in the follow-up docs/benchmark pass.

RAT1 skips corpus inference, but local still loads the model during warm for the identity check. The old debug-build numbers have also been replaced with release-build results.

src/core/src/tool_registry.rs
faustoq: Done. I added a shared `test_support.rs` module and moved the artifact/warm helpers introduced in this PR there (`ArtifactBuildStub`, `PanicOnEmbedStub`, counting/failing embedders, and `unit*` helpers).

There's still some older test scaffolding elsewhere that could use the same module in a follow-up, but I've kept that broader cleanup out of scope for this pass.

faustoq:
Updated after review.

All requested changes and review comments have been addressed.

I also replaced the previous benchmark figures in the PR description with release-build measurements and clarified the scope/limitations of artifact warming.

Ready for final review / approval

src/core/src/embedding_config.rs
```diff
@@ -89,14 +96,300 @@ pub(crate) fn huggingface_fingerprint(repo: &str, revision: &str) -> String {
...
+    paths: &[&Path],
+    expected_stamps: Option<&[FileStamp]>,
+) -> Result<String, EmbedderError> {
+    let name = dir.display().to_string();
+    let stamps: Vec<FileStamp> = paths
+        .iter()
+        .map(|p| stamp_file(p, &name))
+        .collect::<Result<_, _>>()?;
+
+    if let Some(expected) = expected_stamps
```
rstagi: **[nit] byte-identical touch of a model file bricks artifacts for the process**

The load-time `(len, mtime)` stamp is treated as authoritative and any drift hard-errors. mtime moves without bytes changing constantly: container re-extract, `cp` without `-p`, rsync, volume remount.

Repro: build an artifact from a local model dir, register a semantic catalog so the embedder is resident, then `utimesSync(dir + "/config.json", ...)` with identical content. Build and warm both fail with `"local model files changed since load ..."`, and retrying in the same process keeps failing. The error says reload the embedder, but the embedder cache is path-keyed and never reloads, so the only remedy is a process restart. Plain semantic search is unaffected.

ADR-0012 calls `(len, mtime)` "a process-local accelerator ... not authoritative" and cites Docker remounts as the reason the content digest exists. Can this invalidate the memo and recompute instead?

src/sdk/ts/src/catalog.ts
```diff
@@ -295,6 +310,15 @@ export class ToolCatalog {
       }
       this.tools.set(tool.id, metadata);
     }
+    await this.ensureDenseReady();
+  }
+
+  private async ensureDenseReady(): Promise<void> {
+    if (this.embeddingArtifact) {
+      const { bytes, onMiss } = await resolveEmbeddingArtifact(this.embeddingArtifact);
```
rstagi: **[minor] the artifact read yields before the dense permit is claimed, so "registry busy" never fires**

The dense permit is claimed in the synchronous NAPI body, so its timing depends on when JS calls the native method. `buildDense()` gets called immediately, but the artifact branch `await`s `resolveEmbeddingArtifact` first, which yields on `fs.readFile` with the permit still at 0. A second `register`/`replaceAll` slips into that window and mutates the corpus unchallenged:

```
plain    : [ 'fulfilled', 'registry busy; await the active operation bef...' ]
artifact : [ 'fulfilled', 'fulfilled' ]
```

No half-warmed cache, since the warms still serialize on the dense gate. What breaks is diagnostics and the documented contract. If the artifact covers only tool A and B lands in the window, `register(A)` fails with `Incomplete missing=["B"]`, complaining about a tool it never registered. And for `replaceAll`, whose doc-comment promises it "refuses the call outright rather than applying it half-way", two interleaved calls both succeed and the first one's supposedly-final counts describe a corpus that got replaced before any embedding ran.

Same at `skill-catalog.ts:163`, which matters more since `replaceAll` is the documented reload path. Reading the bytes before the corpus mutation, or claiming the permit before the read, should cover it.

src/core/src/dense_cache.rs
```diff
@@ -164,6 +224,148 @@ impl DenseCache {
...
+        let mut active_artifact = embedder.artifact_identity()?;
+        let mut runtime_identity = embedder.fingerprint();
+        if matches!(self.model, EmbeddingModel::Endpoint { .. })
+            && active_artifact != header.model_fingerprint
+        {
+            let probed = embedder.embed_batch_with_identity(&[ARTIFACT_WARM_PROBE_TEXT.into()])?;
+            active_artifact = probed.fingerprint.clone();
+            runtime_identity = probed.fingerprint;
+        }
+        if active_artifact != header.model_fingerprint {
```
rstagi, line 317: **[nit] mismatch message points users at the wrong fix**

`EmbedderError::ModelMismatch` was worded for a different case. Built an artifact with a local model, warmed with the HF default:

```
embedding model mismatch: cache was built with local|content=64:fb6bd45e36cf908a...,
active model is hf|repo=..., (hint: the configured model changed; re-embed the corpus with the new model)
```

On the likeliest failure here (swap the model, forget to rebuild):

- `built` is the artifact fingerprint, not the cache's. Runtime Local identity is path-based and never `content=`, so "cache was built with" describes an impossible state.
- the fix is rebuild the artifact, not re-embed the corpus.
- `content=64:<sha256>` is in no user-visible config, and the artifact path is never named.

A variant carrying the artifact identity plus a rebuild hint would make this self-service.

src/sdk/ts/src/registry.ts
```diff
@@ -99,6 +99,43 @@ export class ToolRegistry {
...
+    }
+  }
+
+  /**
+   * Warm the dense cache from a build-time embedding artifact (ADR-0018).
+   * `onMiss` is `"error"` (fail with {@link ArtifactWarmError} / `"Incomplete"`
+   * when some corpus ids are uncovered) or `"embed"` (embed only the missing
+   * ids). Independent of the registry's search method.
+   */
+  async experimentalWarmEmbeddingsFromArtifact(
```
rstagi, line 130: **[nit] warm skips the model-mismatch warning buildDense emits**

`buildDense()` ends with `#maybeWarnModelMismatch()`, this does not, and `ensureDenseReady()` returns early so it never reaches `buildDense()`. A catalog using an artifact plus `experimentalEnableAdaptiveRanking` loses the warning. Same for the `SkillRegistry` copy at 395. Reads as unintentional.

rstagi (approved):
lgtm 🚀 left a couple of minor comments, feel free to tackle them immediately, tackle after we merge this, or file them as issues... your call :)

src/core/src/embedding_config.rs
faustoq: Good catch! Fixed. Stamp drift now recomputes and compares the established content digest, so the byte-identical metadata changes are accepted, while the real content drift still fails closed. 

The error message now points to a fresh process instead of an in-process reload.

src/sdk/ts/src/catalog.ts
faustoq: Fixed. Artifact resolution now runs under the registry-busy guard in both SDKs, so racing mutations are rejected before they can alter the corpus.

PR #149, opened 2026-08-13
feat: runtime events lane — core stream + TS/Python SDK seams

rstagi:
## Goal

Give the Ratel SDK a first-class, subscribable runtime-events stream — Rust core plus TS and Python seams — with stable IDs shared with the OTel projection, so `@ratel-ai/cloud-sdk` can attach as a consumer and Ratel Cloud stops depending on OTel sampling/content-capture for product facts. OTel emission stays unchanged (parallel projection, never rebased on the stream).

Plan: [Runtime events lane — Multi-Phase Plan](https://app.kestral.ai/workspace/10V6J6nL/documents/1IEI1bL0) · Effort task: [RS-79](https://app.kestral.ai/workspace/10V6J6nL/task/1IENAQuT)

## Phases (all merged; per-phase verify + effort verify green at each merge)

1. **ADR: runtime-events lane** + amend ADR-0007/0003/0014, CONVENTIONS.md `ratel.event.id` — [RS-80](https://app.kestral.ai/workspace/10V6J6nL/task/1IEQQV0e) · `694409e4`
2. **Envelope v2 stable IDs + fan-out sink** (`event_id`/`invocation_id`/`source_id`, bounded queues, drop-oldest, `events_dropped`) — [RS-81](https://app.kestral.ai/workspace/10V6J6nL/task/1IERL63v) · `f3b11514`
3. **FFI push bridge** — napi ThreadsafeFunction (TS) + pyo3 callback sink (Python), batched off-hot-path — [RS-82](https://app.kestral.ai/workspace/10V6J6nL/task/1IESDCyQ) · `2beba041`
4. **TS SDK: `events.subscribe` + `catalog.snapshot`**, `ratel.event.id` stamped on OTel projection, composite experiment sink, conformance fixtures — [RS-83](https://app.kestral.ai/workspace/10V6J6nL/task/1IET6c5i) · `d8dd3d5e`
5. **Python SDK mirror** — same subscribe/snapshot seam, `call_soon_threadsafe` marshaling — [RS-84](https://app.kestral.ai/workspace/10V6J6nL/task/1IETxXLU) · `fb11a25b`

## Progress digest

- Integration branch cut from `main` @ `e7d2933a`; 5 lane branches merged serially, effort verify (`cargo test --workspace && pnpm --filter @ratel-ai/sdk test`) green at each merge.
- Phase 5 merge initially reported a semantic-conflict false alarm (fresh integration venv lacked mypy); env fixed, typing tests pass, merge intact.
- Loop run id: `loop-runtime-events-sdk-2026-08-13-171939` (5 phases, 0 HIL pauses).

### DIFF
--- a/src/sdk/ts/src/runtime-events.ts
+++ b/src/sdk/ts/src/runtime-events.ts
@@ -1,0 +1,179 @@
+import { randomBytes, randomUUID } from "node:crypto";
+import type { NativeEventSubscription } from "../native/index.cjs";
+import type { ToolDefinition } from "./catalog.js";
+import type { SkillDefinition } from "./skill-catalog.js";
+
+const DEFAULT_SOURCE_ID = "ratel";
+
+/** Stable envelope shared by every public runtime event (ADR-0019). */
+export interface RuntimeEvent {
+  readonly v: 2;
+  readonly event_id: string;
+  readonly ts: number;
+  readonly session_id: string;
+  readonly source_id: string;
+  readonly type: string;
+  readonly invocation_id?: string;
+  readonly catalog_version?: string;
+  readonly environment?: string;
+  readonly end_user_id?: string;
+  readonly trace_id?: string;
+  readonly span_id?: string;
+  readonly [field: string]: unknown;
+}
+
+/** Identity and bounded-delivery options for {@link RuntimeEvents}. */
+export interface RuntimeEventsOptions {
+  /** Stable identity for this process/session. Defaults to a fresh UUID. */
+  sessionId?: string;
+  /** Stable deployment source. Defaults to OTel `service.name`, then `"ratel"`. */
+  sourceId?: string;
+  /** Per-registry subscriber queue capacity. Defaults to the native bridge default. */
+  queueCapacity?: number;
+  /** Maximum envelopes per callback batch. Defaults to the native bridge default. */
+  batchSize?: number;
+}
+
+/** One runtime-events subscription. */
+export interface RuntimeEventSubscription {
+  /** Stop accepting future events for this subscriber. */
+  unsubscribe(): void;
+  /** Wait until work already accepted by both registry streams reaches the handler. */
+  flush(): Promise<void>;
+  /** Envelopes dropped from this subscriber's bounded native queues. */
+  readonly droppedCount: number;
+}
+
+/** Complete executor-free catalog state published separately from runtime events. */
+export interface CatalogSnapshot {
+  readonly sourceId: string;
+  readonly tools: readonly ToolDefinition[];
+  readonly skills: readonly SkillDefinition[];
+}
+
+/** Public catalog-state seam. */
+export interface RuntimeCatalog {
+  /** Return a current, serializable full replacement snapshot. */
+  snapshot(): CatalogSnapshot;
+}
+
+type RuntimeEventHandler = (
+  batch: readonly RuntimeEvent[],
+) => void | PromiseLike<void>;
+
+interface EventSource {
+  subscribeEvents(
+    handler: (batch: RuntimeEvent[]) => void,
+    options: Required<RuntimeEventsOptions>,
+  ): NativeEventSubscription;
+}
+
+/** Public merged push stream over tool, skill, and SDK-owned runtime facts. */
+export class RuntimeEvents {
+  readonly sessionId: string;
+  readonly sourceId: string;
+  readonly #options: Required<RuntimeEventsOptions>;
+  readonly #sources: readonly EventSource[];
+  readonly #sdkSubscribers = new Set<RuntimeEventHandler>();
+
+  /** @internal Constructed by {@link ratel}; consumers use `runtime.events`. */
+  constructor(sources: readonly EventSource[], options: RuntimeEventsOptions = {}) {
+    this.sessionId = options.sessionId ?? randomUUID();
+    this.sourceId = options.sourceId ?? defaultSourceId();
+    this.#options = {
+      sessionId: this.sessionId,
+      sourceId: this.sourceId,
+      queueCapacity: options.queueCapacity ?? 1_024,
+      batchSize: options.batchSize ?? 64,
+    };
+    this.#sources = sources;
+  }
+
+  /**
+   * Subscribe to merged runtime-event batches. Delivery is asynchronous and
+   * bounded; handler work never blocks the emitting registry operation.
+   */
+  subscribe(handler: RuntimeEventHandler): RuntimeEventSubscription {
+    const deliver = (batch: RuntimeEvent[]): void => {
+      try {
+        void Promise.resolve(handler(batch)).catch(() => {});
+      } catch {
+        // Runtime-event consumers are observational and fail open.
+      }
+    };
+    const subscriptions = this.#sources.map((source) =>
+      source.subscribeEvents(deliver, this.#options),
+    );
+    this.#sdkSubscribers.add(handler);
+    let active = true;
+    return {
+      unsubscribe: () => {
+        if (!active) return;
+        active = false;
+        this.#sdkSubscribers.delete(handler);
+        for (const subscription of subscriptions) subscription.unsubscribe();
+      },
+      flush: async () => {
+        await Promise.all(subscriptions.map((subscription) => subscription.flush()));
+      },
+      get droppedCount() {
+        return subscriptions.reduce((total, subscription) => total + subscription.droppedCount, 0);
+      },
+    };
+  }
+
+  /** @internal Merge an SDK-owned fact (experiments) into every public subscriber. */
+  emit(event: Record<string, unknown>): RuntimeEvent {
+    const envelope: RuntimeEvent = {
+      v: 2,
+      event_id: newUlid(),
+      ts: Date.now(),
+      session_id: this.sessionId,
+      source_id: this.sourceId,
+      type: String(event.type ?? "unknown"),
+      ...event,
+    };
+    for (const handler of this.#sdkSubscribers) {
+      queueMicrotask(() => {
+        try {
+          void Promise.resolve(handler([envelope])).catch(() => {});
+        } catch {
+          // Runtime-event consumers are observational and fail open.
+        }
+      });
+    }
+    return envelope;
+  }
+}
+
+function defaultSourceId(): string {
+  if (process.env.OTEL_SERVICE_NAME) return process.env.OTEL_SERVICE_NAME;
+  const serviceName = process.env.OTEL_RESOURCE_ATTRIBUTES?.split(",")
+    .map((entry) => entry.trim().split("=", 2))
+    .find(([key]) => key === "service.name")?.[1];
+  return serviceName || DEFAULT_SOURCE_ID;
+}
+
+/** Minimal monotonicity-free ULID generator: event uniqueness, time sorting, wire alphabet. */
+function newUlid(now = Date.now()): string {
+  const alphabet = "0123456789ABCDEFGHJKMNPQRSTVWXYZ";
+  let time = now;
+  let encodedTime = "";
+  for (let index = 0; index < 10; index += 1) {
+    encodedTime = alphabet[time % 32] + encodedTime;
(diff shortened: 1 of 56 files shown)


rstagi:
# Review: PR #149 — feat: runtime events lane (core stream + TS/Python SDK seams)

**Verdict: request changes.** The lane itself is well designed and well tested, and everything I ran is green. Two majors need resolving before merge: an open TS subscription pins the Node process forever (demonstrated below), and this PR silently re-ships the entire retrieval-experiment engine that is still open for review as PR #140. Plus 4 minors and 2 nits. The stacked-split question has a concrete answer at the end.

Reviewed at head `fb11a25b` in a detached worktree.

## Step 1: built and ran it

All of CI's gate, run locally, passes:

- `cargo build --workspace`, `cargo test --workspace` (includes 20+ new trace/fanout tests), `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --check --all`: clean.
- `pnpm -r build / typecheck / lint / test`: all packages pass (native module rebuilt, native-events tests exercise the real .node binary).
- Python SDK: `ruff` clean, `pytest` 250 passed, 2 skipped. (`mypy` fails under the bare AGENTS.md recipe because `telemetry.py` imports `opentelemetry`; pre-existing on `main`, CI installs `opentelemetry-sdk` + the telemetry package so `ci-gate` is unaffected.)

Live smoke, both languages: register tool → search → invoke with an active `events.subscribe`. Envelopes arrive batched with `v:2`, ULID `event_id`, shared `session_id`/`source_id`, `invoke_start`/`invoke_end` paired by one `invocation_id`; `catalog.snapshot()` returns sorted definitions with executors verifiably absent; `flush()` and `droppedCount` behave. The core fan-out (drop-oldest, loss windows, `events_dropped`, non-blocking emit, close-drains-queue) reads carefully and its condvar predicates are actually disjoint; the stalled-JS drop test proves the bound end to end.

## Findings

**[major] an open TS subscription keeps the Node process alive forever** — `src/sdk/ts/src/runtime-events.ts:163` / `src/sdk/ts/native/src/lib.rs:268`. The napi threadsafe callback is built strong (`ThreadsafeFunction<..., false, 1>`, Weak=false), so the event loop never drains while a subscription exists. Repro: the smoke script above with `sub.unsubscribe()` exits in ~1s; the identical script without it never exits (killed by a 15s timeout, repeatable). Python is unaffected (same script minus unsubscribe exits cleanly). For a fail-open observational stream this fails silent: any CLI, batch job, or test runner that attaches a consumer and forgets teardown hangs at exit, and neither README nor ADR-0019 mentions it. Long-running servers are unaffected. Fix options: build the TSFN weak so it never pins the loop, or expose `unref()` and document the lifecycle; a weak default matches "subscriber code can never block the emitting operation" in spirit.

**[major] re-ships open PR #140 wholesale** — `src/sdk/ts/src/experiment*.ts`, `src/telemetry/*`. Roughly 4.5k of the 11.5k added lines are PR #140's retrieval-experiment engine, re-authored as new commits (e.g. "refactor(sdk): split experiment.ts into focused modules" is `73252f96` on #140 and `60a7e58e` here; `git diff` between the two heads over `experiment*.ts` is only the event-ID join, ~300 lines). PR #141 stacks on #140. Whichever merges first makes the other conflict or become review noise, and #140's review comments silently never apply to the copy that lands. No code defect; this needs an explicit decision recorded on both PRs: either rebase this PR onto #140 (making it a true stack) or close #140/#141 as absorbed and port their open review threads here.

**[minor] `flush()` semantics diverge between TS and Python** — `src/sdk/ts/src/runtime-events.ts:166` vs `src/sdk/python/ratel_ai/runtime_events.py:101`. Python's `flush()` awaits async handler completion (pending futures are tracked and drained). TS tracks pending promises only for SDK-emitted experiment events; native batches are fire-and-forget (`void Promise.resolve(handler(...)).catch(() => {})`), so TS `flush()` resolves once the handler was invoked, not when its promise settles. A TS consumer doing flush-then-shutdown (exactly the cloud-publisher shape this lane exists for) can lose in-flight async work that the Python twin would deliver. Either track handler promises in TS or state the weaker TS contract in both READMEs.

**[minor] `ratel.event.id` is missing from the vocabulary source of truth** — `src/telemetry/core/src/lib.rs`. TS (`src/telemetry/ts/src/index.ts:110`) and Python (`ratel_ai_telemetry/__init__.py:90`) both export `RATEL_EVENT_ID`, CONVENTIONS.md documents it, but the Rust constants crate has no such constant and `conformance/fixtures.json` doesn't pin the attribute key. On top of that the TS SDK hardcodes its own local `const RATEL_EVENT_ID = "ratel.event.id"` (`src/sdk/ts/src/telemetry.ts:126`) instead of importing from `@ratel-ai/telemetry`, while the Python SDK imports it properly. This is exactly the cross-language drift the conformance machinery exists to prevent; add the Rust constant, pin the key in fixtures, and import in the TS SDK.

**[minor] ADR-0019 requires facts the implementation doesn't emit** — `docs/adr/0019-runtime-events-lane.md:36`. The v1 table lists `search_id` as a required product fact for search events and "ordered `hits[]` of target id, rank, and score". Nothing emits `search_id` (the only other occurrence in the repo is a fabricated field in a truncation test), and hits carry id + score with rank implied by order. Either amend the ADR (envelope `event_id` identifies the search; rank is positional) or implement the fields. As written, a Cloud receiver built from the ADR would reject or mis-model every search event.

**[minor] Python bridge drops a whole batch when one envelope fails to convert** — `src/sdk/python/native/src/lib.rs:283`. `pythonize(...)?` inside the batch loop aborts the `with_gil` closure, and `let _ =` swallows it, so one bad envelope silently discards up to `batch_size` events with no `events_dropped` accounting. The TS bridge (`src/sdk/ts/native/src/lib.rs:319`) filter-maps per envelope. Unlikely with serde-owned types, but the two bridges should fail identically; per-envelope skip is one line.

**[nit] `RuntimeEvents.emit` spread order lets callers clobber envelope identity** — `src/sdk/ts/src/runtime-events.ts:196`. `...event` comes after the defaults, so `type: String(event.type ?? "unknown")` is dead whenever `event.type` is set, and a caller can silently override `v`, `ts`, or `session_id`. The `event_id` override is intentional (shared-ID join); the rest deserve either explicit picks or a comment saying overriding is allowed.

**[nit] `unsubscribe()` doc overpromises** — `src/sdk/ts/native/src/lib.rs:94` (and the pyo3 twin). "Stop future delivery" is not quite what happens: the core dispatcher drains already-queued envelopes to the callback after close, by design (close-drains is even what makes flush-after-unsubscribe coherent). Wording fix only.

No AI slop found: comments carry contract content, the module split is real, and the giant constant tables are the vocabulary itself. Test quality is genuinely good, including the concurrency pairing test, the stalled-JS drop-oldest proof, cross-language conformance fixtures, and type-tests.

## Stacked-PR split

Signals are all present: 11.5k additions, five independently verified units (effort verify was green at each phase merge, so the seams are proven), and a third of the diff duplicates an open PR. Reviewing this as one unit is what allowed the #140 collision to go unnoticed. Recommended stack, plain `gh` branches, each base = previous:

1. `docs/runtime-events-adr` (base `main`): commit `b0df24ae` — ADR-0019, ADR-0003/0007/0014 amendments, CONVENTIONS.md, telemetry README.
2. `core-envelope-v2-fanout` (base 1): `25bc415f`, `480a4718`, `6526e54d` — `src/core/` envelope v2 + fan-out sink + tests.
3. `ffi-event-push-bridges` (base 2): `f585f06b`, `d521092d`, `dab67d66`, `ba8c8b25` — both `native/src/lib.rs` bridges.
4. Retrieval experiments = **existing PR #140 rebased onto 3** (or closed-as-absorbed with its review threads ported): `c2dc31a4` through `60a7e58e`, `512b2672` — `experiment*.ts`, telemetry vocabulary.
5. `ts-events-subscribe` (base 4): `a30b84e6`, `f815496e`, `d3cb977f` — `runtime-events.ts`, `catalog.snapshot`, OTel ID join, conformance fixtures.
6. `py-events-subscribe` (base 5): `cb8f5add`, `76d04193`, `13c8713f`, `c9459e59`, `7e925b70`, `d01b0e68` — Python mirror.

If the team prefers to land #149 as one PR anyway (the per-phase verify history is a reasonable argument), the #140/#141 decision still has to happen first; that part is not optional.

src/sdk/ts/src/runtime-events.ts
```diff
@@ -0,0 +1,369 @@
...
+    bits += 8;
+    while (bits >= 5) {
+      bits -= 5;
+      encodedEntropy += alphabet[(buffer >> bits) & 31];
+    }
+  }
+  return encodedTime + encodedEntropy;
+}
+
+function normalizeRuntimeEvent(input: Record<string, unknown>): RuntimeEvent {
```
kayraucklnc, line 278: **Perf: every event is fully serialized at least twice, and it lands on the Node main loop.**

`normalizeRuntimeEvent` runs per event inside the threadsafe-function callback, so this is main-thread work. It deep-clones the event via `sanitizeValue`, then calls `serializedSize` — a full `JSON.stringify` — purely to compare against the 64 KB cap. The Cloud publisher then stringifies the same object *again* in `createBatches` (ratel-cloud-sdk#6, `publisher.ts:235`).

The oversize branch is worse: it re-runs `serializedSize` after each deleted key, so a large event costs O(k) full serializations to trim.

Suggestion: do a cheap lower-bound check first (the native side already knows the envelope size before it crosses the bridge), and pass the serialized string through to the consumer so the batcher doesn't redo the work. Not a blocker — but this is the one piece of per-event cost that sits on the main loop rather than a background thread.

kayraucklnc (commented):
Reviewed this together with ratel-cloud-sdk#6 and ratel-cloud#130 — checked out into a worktree, built, and ran `cargo test --workspace` plus the TS SDK suite. All green (Rust workspace + doctests, 557 TS tests).

The design holds up well. ADR-0019 is unusually clear about what it is *not* doing, the `event_id` → `ratel.event.id` join is the right primitive for keeping the two lanes from double-counting, and the fan-out sink in `core/src/trace/sink.rs` is genuinely careful concurrency: bounded queues, drop-oldest with a reported loss window, `catch_unwind` around subscriber code, and a dispatcher a slow consumer can't stall.

I also confirmed the zero-cost claim for people who never subscribe: `event_stream` starts as `None`, `active_trace_sink` falls through to the base sink, and the registry default is `NoopSink`. No fan-out, no threads, no serialization. Local-only users pay nothing for this existing, which is the important property.

One perf note inline. Everything else I found lives in the other two PRs.

**Build note (not caused by this PR):** `pnpm --filter @ratel-ai/sdk test` fails on a clean checkout because `@ratel-ai/telemetry` resolves to `dist/` and isn't built first. Pre-existing monorepo ordering, but it does mean the verify command in the description needs `pnpm --filter @ratel-ai/telemetry build` in front of it to reproduce.

src/sdk/ts/src/runtime-events.ts
rstagi: Fixed in 396d5bec — `serializedSizeUpperBound` short-circuits the full stringify for small events, and the oversize branch now sizes incrementally per removed key instead of re-serializing each pass. Passing the pre-serialized envelope through to the batcher is deferred as a cross-repo design change (comment in code); the batcher half is fixed independently in ratel-cloud-sdk 6c60162 (single stringify reused for size + body). Your build note landed too: 98257e5a prefixes the verify command with `pnpm --filter @ratel-ai/telemetry build`.

kayraucklnc (approved):
Approving. Re-reviewed at 98257e5 in a clean worktree and re-ran the gates.

One note so it isn't mistaken for a regression: `pnpm -r typecheck` fails in `src/adapters/ts-vercel-ai-sdk` (`aisdk.ts:194`, `:305`). That's pre-existing — main fails the same check with 7 errors, this branch with 2, and the PR touches no file under `src/adapters/`. The SDK type changes here actually improve inference there. Worth fixing separately.

Thanks for the README fix on the telemetry build ordering too.

PR #150, opened 2026-08-14
ci: adapter sdk peer range

faustoq:
## Problem

Both framework adapters publish their `@ratel-ai/sdk` peer as `^<in-repo SDK version>`.
At 0.x a caret pins the **minor** (`^0.6.0` means `>=0.6.0 <0.7.0`), so every SDK minor orphans both adapters.
`npm install @ratel-ai/mastra @ratel-ai/sdk@latest` fails with `ERESOLVE`. Broken since SDK 0.7.0 (2026-08-07).

The daily `verify-install` cron caught it a few times (2026-08-08 -> 2026-08-13), then went green because #148 removed the SDK co-install from the adapter verify jobs. The check stopped exercising the failure mode but the breakage still exists.

## Solution

Publish a floor range instead of a caret: `SDK_ADAPTER_PEER_FLOOR` in `scripts/release-units.mjs` is the single source of truth. `scripts/pin-adapter-sdk-peer.mjs` applies it in both `release.yml` publish jobs and in `publish-rc.sh`.
Source manifests stay `workspace:^`.

The floor is `0.9.1`: vercel's shipped code needs `ExperimentalPassthroughToolExposure`, first exported in SDK 0.9.1 (verified 94/94 and 52/52 against 0.9.1). `>=0.6.0` would resolve and then fail on a missing export. ADR-0020 records the measurement.

The SDK publish-order `npm view` guard is dropped: the peer no longer names the in-repo SDK version, so a floor-range check could never fail. Telemetry pins and their guards are unchanged.

## Notes

- Takes effect on the next adapter release (`vercel-ai-sdk@0.4.0`, `mastra@0.3.0`). This PR does not change any published package by itself.
- `release.yml` added to the `ts` paths filter, so a release.yml-only PR still runs the contract tests that guard the pin
- `publish-rc.sh` restores the manifests through a guarded `git checkout` EXIT trap. A core/telemetry-only run can't clobber an unrelated uncommitted edit
- The mechanism is 0.x-only: at SDK 1.0.0, `^1.2.0` already means `>=1.2.0 <2.0.0`, so we can delete it and let `release.yml` go back to a plain caret. Explained in ADR-0020.
- After both adapters ship, we can restore the `sdk@latest` co-install to `verify-install.yml` so the cron catches this class of drift again. It also needs the `workflow-contracts` "does not interpolate the SDK version" assertion updated.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,6 +99,7 @@ jobs:
               - 'Cargo.lock'
               - 'rust-toolchain.toml'
               - '.github/workflows/ci.yml'
+              - '.github/workflows/release.yml'
             # maturin builds the Python native binding from the Rust core; the SDK job
             # also installs the telemetry sibling and exercises the pydantic-ai example.
             python:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -572,21 +572,15 @@ jobs:
       - name: Build TS
         run: pnpm --filter "@ratel-ai/mastra..." run build
       # npm keeps `workspace:` specifiers verbatim, so pin the @ratel-ai/sdk peer AND
-      # dev deps to the concrete co-released version before publishing (ephemeral CI
+      # dev deps to the ADR-0020 floor range before publishing (ephemeral CI
       # edit; not committed). Keeps the npm-OIDC dir publish + provenance identical to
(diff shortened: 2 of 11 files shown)


rstagi (approved):
lgtm!

PR #152, opened 2026-08-15
feat: attach definition overrides through a vendor-neutral overlay seam

rstagi:
Stack 3/3; depends on #155.

Adds the experimental, vendor-neutral definition-overlay seam. An injected `ExperimentalDefinitionOverlaySource` supplies retrieval-description overrides through `catalog.experimentalAttachDefinitionOverrides({ source })`; the method call itself is the one-way process opt-in, so there is no disabled/no-op boolean form.

All supporting public types use the `Experimental*` prefix. The SDK still owns no URL, credential, or vendor behavior; `@ratel-ai/cloud-sdk` is simply the first source implementation. Wire format, ETag/304 refresh, override-wins precedence, warn-once shadowing, local-only snapshots, and the neutral `ratel.catalog.use_definition_overrides` telemetry attribute are preserved.

Stack: #154 → #155 → #152.


src/sdk/ts/src/telemetry.ts
```diff
@@ -181,6 +193,95 @@ export interface RuntimeEventProjection {
...
+function canonicalJson(value: unknown): string {
+  return JSON.stringify(sortJson(value));
+}
+
+function sortJson(value: unknown): unknown {
+  if (Array.isArray(value)) return value.map(sortJson);
+  if (value !== null && typeof value === "object") {
+    return Object.fromEntries(
+      Object.entries(value as Record<string, unknown>)
+        .sort(([left], [right]) => left.localeCompare(right))
```
rstagi: [major] the TS content_hash diverges from the Rust core and Python for the same definition

`localeCompare` is locale-aware ICU collation, but the Rust core sorts keys by byte order (serde_json `sort_keys`) and Python by code point (`json.dumps(sort_keys=True)`), which agree with each other. I verified the Rust test pin `b114cd9a...` reproduces exactly from Python's canonicalization. TS disagrees whenever collation reorders keys, e.g. mixed case, and the result also depends on the host locale.

End-to-end repro (one process, one registration with `inputSchema: {properties: {a, B}}`):

```
Rust core JSONL stream: 39f64455d6bee498a4638cf45a3f0549e3261c0ccf7758387c92a1731bd32baf
TS OTel event:          6a7eb62c1ed761ee544e56e7dd46c2deac0fd560285fbed57a7bea5a557b899d
```

Python canonicalization of the same content matches the core. Any consumer joining or deduping across the local stream and OTLP on `ratel.catalog.content_hash` silently mismatches; all-lowercase-key definitions are unaffected, which is why the pinned `a6135789...` test passes in both TS and Python. Fix: sort by code-unit order, `.sort(([l], [r]) => (l < r ? -1 : l > r ? 1 : 0))`.

protocol/v2/conformance/README.md
```diff
@@ -0,0 +1,51 @@
+# `protocol/v2/conformance/`
```
rstagi, line 1: [minor] CI never runs this verifier

The `protocol` job in `.github/workflows/ci.yml` (line 492, untouched by this PR) runs only `node protocol/v1/conformance/verify.mjs`. The v2 suite passes locally (47 checks) but nothing gates it, so the newly frozen eight-field ETag projection can drift silently. The `protocol/**` path filter already triggers the job for v2 edits; it just checks nothing of v2. One-line fix: add a v2 verify step.

src/telemetry/conformance/fixtures.json
```diff
@@ -588,6 +588,125 @@
...
+            "ratel_catalog_kind": "tool",
+            "ratel_catalog_id": "read_file",
+            "ratel_catalog_name": "read_file",
+            "ratel_catalog_description": "Read files",
+            "ratel_catalog_tags": [],
+            "ratel_catalog_input_schema": "{\"type\":\"object\"}",
+            "ratel_catalog_output_schema": "{\"type\":\"string\"}",
+            "ratel_catalog_searchable_description": "open documents",
+            "ratel_catalog_searchable_description_overridden": true,
+            "ratel_catalog_content_hash": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
```
rstagi, line 744: [minor] no shared cross-language vector pins content_hash

This fixture uses placeholder hashes, the TS and Python tests pin the same all-lowercase fixture, and the Rust test pins a different one, so no single definition has its hash asserted identical across all three emitters. A shared vector with a mixed-case schema key would have caught the localeCompare divergence flagged in telemetry.ts. Suggest one real pinned hash here, asserted by all three conformance tests.

src/sdk/ts/src/registry.ts
```diff
@@ -377,7 +381,9 @@ export class SkillRegistry {
    */
   replaceAllItems(items: readonly Skill[]): ReplaceOutcome {
     assertNotArtifactBusy(this);
-    return this.native.replaceAll([...items]);
+    const outcome = this.native.replaceAll([...items]);
+    recordCatalogDefinitions("skill", items, this.emittedDefinitionHashes);
```
rstagi: [minor] the dedup map never forgets removed definitions

`emittedDefinitionHashes` only accumulates. After `replaceAll` drops a skill id, re-registering the identical skill later emits `catalog_definition` on the core stream (the core compares against the current corpus, which no longer holds the id) but this channel suppresses it because the stale hash is still mapped. The two streams disagree about the re-add. Pruning entries for ids absent from the `replaceAll` batch restores parity.

src/core/src/tool_registry.rs
```diff
@@ -410,6 +411,11 @@ impl ToolRegistry {
     /// ```
     pub fn register(&mut self, tool: Tool) {
         let tool_id = tool.id.clone();
+        let definition = TraceEvent::catalog_definition_for_tool(&tool);
+        let definition_changed = self.tools.get(&tool_id).is_none_or(|existing| {
```
rstagi: [nit] every register builds and hashes two full definitions even when nothing changed

Change detection constructs a complete `CatalogDefinition` for both the incoming and the existing entry (cloning both schemas) and runs canonicalization plus SHA-256 twice per call, including when the sink is `NoopSink`. Same pattern in the skill and fact registries and in `replace_all`. Correct, just wasted work on the hot re-register path; comparing the relevant fields directly, or caching the last hash per id, avoids it.

src/sdk/ts/native/src/lib.rs
```diff
@@ -833,11 +833,13 @@ pub struct Tool {
     pub name: String,
     /// What the tool does and when to use it — the main ranking signal.
     pub description: String,
-    /// JSON Schema of the arguments. Property names and their `description`s
-    /// (nested included) are indexed for ranking.
+    /// Optional replacement for the description component used by retrieval.
+    /// The tool name stays indexed; schemas are never indexed.
+    pub searchable_description: Option<String>,
```
rstagi: [nit] `searchableDescription: null` is rejected at the NAPI boundary

Untyped JS passing `null` (protocol v2's canonical unset spelling) fails loud with "Failed to convert JavaScript value Null into rust type String" (verified against the built binding). Fine since it fails loud, but a future v2 catalog loader must strip wire nulls before forwarding skills into the SDK; worth a `?? undefined` normalization at ingest or a note.

rstagi (commented):
Review of head d6ef7aba, run in a detached worktree.

**Verdict:** solid PR; the implementation matches ADR-0021 exactly (override swaps only the description component, name and tags stay indexed, schemas dropped from ranking) and the docs are consistent. One cross-language defect should be fixed before merge (TS content_hash diverges from core/Python, demonstrated end to end), plus a CI gap for the v2 conformance vectors. Details inline.

**Validation:** cargo fmt/clippy/test workspace PASS (357 core tests). pnpm -r build/typecheck/lint/test PASS. Python SDK ruff/mypy/pytest PASS (302 passed, 1 skipped). `node protocol/v2/conformance/verify.mjs` PASS (47 checks). Live-exercised the built TS SDK: `ratel.catalog.definition` emits, dedups identical re-registers, and the core JSONL stream carries the matching event.

**Note:** the PR body still shows only Phase 3 merged; the branch also contains Phases 1, 2, and 4. Worth updating before merge.

**Stacked-PR split:** none recommended. The phase merge commits already provide dependency-ordered review seams (protocol v2 -> core -> SDK/adapters -> telemetry events); splitting now would only add churn.

:)

rstagi:
Review findings addressed in 57ad07b4:
- [major] fixed: `sortJson` now sorts by code units (matches Rust/Python); mixed-case cross-language vector pinned (`03542abe…`, computed via the Python twin). SDK suite 574/574 green.
- [minor] fixed: CI now runs the protocol v2 conformance verifier (47 checks pass locally).

Left as follow-ups (reviewer minors): shared cross-language content_hash conformance vector; TS dedup map pruning on `replaceAll` removals.

src/sdk/ts/src/catalog.ts
```diff
@@ -402,9 +406,31 @@ export class ToolCatalog {
...
       }
     }
+    await this.registerEffective(
+      batch.map((tool) => this.applyCloudDefinition(tool)),
+      batch,
+    );
+  }
+
+  /** @internal Apply a complete Cloud overlay while retaining local definitions for restore. */
+  async applyCloudDefinitions(overrides: ReadonlyMap<string, string>): Promise<void> {
```
rstagi: [major] snapshot() publishes the Cloud-overridden value as the local definition

`registerEffective` stores the effective (overlay-applied) metadata in `this.tools`, and `snapshot()` reads `this.tools`; `localTools` keeps the local originals but snapshot never uses them. Same shape in `skill-catalog.ts` (`replaceAllEffective` -> `this.skills`). Live run against the built SDK:

```
SNAPSHOT BEFORE attach, deploy.searchableDescription = localcanaryterm
SNAPSHOT AFTER attach,  deploy.searchableDescription = cloudrollbackterm
```

The ts README says "Registration always owns the complete local definition and syncs that definition up", and the sdk README says register-up/attach-down. The snapshot channel has no per-entry cloud marker (unlike the OTel events, which carry the flag), so a Cloud consumer of snapshots silently ingests its own override as the local value and shadow/drift detection is defeated. Fix: snapshot from `localTools`/`localSkills` (and the fact equivalent), or mark ownership per entry in the snapshot. Behind the opt-in flag, hence major not blocker.

src/sdk/ts/src/catalog.ts
```diff
@@ -402,9 +406,31 @@ export class ToolCatalog {
...
+    this.cloudSearchableDescriptions = new Map(overrides);
+    this.registry.setUseCloudDefinitions();
+    await this.registerEffective(
+      [...this.localTools.values()].map((tool) => this.applyCloudDefinition(tool)),
```
rstagi: [minor] every overlay apply re-registers the whole tool corpus and invalidates its embeddings

This re-registers all of `localTools`, and core `register` on an existing id unconditionally runs `bm25.invalidate()` + `dense.invalidate(&tool_id)` even for byte-identical content (`src/core/src/tool_registry.rs:496`; `fact_registry.rs` is identical). Observed live: a refresh returning an identical overlay body still emitted a fresh `index_churn add` on the core stream. For semantic/hybrid catalogs every 200 response, including the empty-overlay initial attach, drops and re-embeds every tool and fact. Skills already avoid this: `replaceAll` invalidates only entries whose name/effective description/tags changed (ADR-0015).

Fix: re-register only entries whose effective searchableDescription actually changes vs the previous overlay. The opt-in OTel re-emit is SDK-level (`recordCatalogDefinitions`) and can stay full. Results stay correct; this is embedding cost, churn noise, and startup latency.

src/sdk/ts/src/telemetry.ts
```diff
@@ -181,6 +194,102 @@ export interface RuntimeEventProjection {
...
+}
+
+function sortJson(value: unknown): unknown {
+  if (Array.isArray(value)) return value.map(sortJson);
+  if (value !== null && typeof value === "object") {
+    return Object.fromEntries(
+      Object.entries(value as Record<string, unknown>)
+        // Code-unit order, never localeCompare: the hash must match the Rust core
+        // (serde_json byte order) and Python (code-point order) on any host locale.
+        .sort(([left], [right]) => (left < right ? -1 : left > right ? 1 : 0))
```
rstagi: [minor] content_hash still diverges from Rust/Python for non-BMP schema keys

The localeCompare fix is right for BMP strings, but JS `<` compares UTF-16 code units while Rust (UTF-8 byte order) and Python (code points) agree with each other. Demonstrated: for keys `"\uFF61"` and `"\u{10000}"`, JS orders `[U+10000, U+FF61]`; Python and Rust order `[U+FF61, U+10000]`. So the comment's "on any host locale" parity claim overreaches: a schema key containing an astral-plane character (e.g. an emoji) alongside a high-BMP key still hashes differently in TS, silently. All-ASCII keys are unaffected.

Fix: compare UTF-8 bytes, e.g. `Buffer.compare(Buffer.from(left, "utf8"), Buffer.from(right, "utf8"))`, or compare code points via `codePointAt` iteration.

src/sdk/ts/src/ratel.ts
```diff
@@ -446,12 +455,52 @@ export function ratel(config: RatelConfig = {}): Ratel {
...
+  };
+  const attachCloudDefinitions = async (
+    options: CloudDefinitionsAttachOptions,
+  ): Promise<CloudDefinitionsAttachment> => {
+    if (!options.useCloudDefinitions) return { refresh: async () => false };
+
+    let etag: string | undefined;
+    const attachment: CloudDefinitionsAttachment = {
+      refresh: async () => {
+        const response = await options.source.fetch(etag);
```
rstagi: [nit] concurrent refresh() calls can apply a stale overlay

Two in-flight refreshes can interleave: the later response applies first, the earlier one then overwrites it, and the stored etag matches neither applied state. The server answers 200 on the mismatched etag so the next refresh self-heals, but until then the stale overlay sits applied. If refresh can ever be driven from more than one place, serialize it with an in-flight promise chain.

src/sdk/ts/src/ratel.test.ts
```diff
@@ -33,6 +33,208 @@ const native = (id: string, description: string): ExecutableTool => ({
...
+      { kind: "tool", entryId: "deploy", searchableDescription: "cloudrollbackterm" },
+    ]);
+
+    await r.catalog.applyCloudDefinitions([]);
+
+    expect(r.tools.search("localcanaryterm", 5).map((hit) => hit.toolId)).toEqual(["deploy"]);
+    expect(r.tools.search("cloudrollbackterm", 5)).toEqual([]);
+  });
+
+  it("applies Cloud retrieval descriptions to skills registered after attach", async () => {
```
rstagi: [nit] skills-after-attach and facts-after-attach are covered, but no test registers a tool after attach; the tool path takes the same `withCloudDefinition` route in `ToolCatalog.register` and is currently only exercised pre-attach.

rstagi (commented):
Re-review (a3) of head 8366376, run in a detached worktree; scope is the four commits since the a1 review at d6ef7ab (content_hash fix + phase 15 useCloudDefinitions attach).

**Verdict:** phase 15 is well built (typed opt-in attach seam, correct ETag discipline, warn-once shadowing matching the docs, opt-in re-emit with `ratel.catalog.use_cloud_definitions=true` exactly as CONVENTIONS.md specifies, constants mirrored across all three telemetry packages + conformance). One [major] to fix before merge: after attach, `catalog.snapshot()` publishes the Cloud-overridden value as the local definition, contradicting the register-up/attach-down ownership model this PR documents (demonstrated live, inline). Two [minor] and three [nit] besides.

**Validation:** `cargo fmt --check` + clippy `-D warnings` + `cargo test --workspace` PASS; `pnpm -r build/typecheck/lint/test` PASS; live exercise of the built SDK (attach with fake overlay source, JSONL core trace) verified search flip, restore on override removal, warn-once, ETag round-trip, and change-sensitive core-stream dedup. No terraform changes in the new work.

**a1 follow-ups:** content_hash localeCompare major fixed (57ad07b); v2 conformance now gated in CI; the cross-language hash pin is TS-test-only (fixture still placeholder); the replaceAll dedup-map minor is still open but not worsened. [nit] the PR body is still stale: only Phase 3 is ticked and phase 15 isn't listed.

No stacked-PR split recommended (unchanged from a1): integration PR of a phased run, merge commits already provide dependency-ordered seams.

:)

rstagi (commented):
## Re-review a4: phase 19 fix verification (targeted)

Scope: c7a0c83 `fix(sdk): preserve local catalog ownership`, closing the two a3 findings. Both are closed, verified live against the built SDK at head 03211ca; no new findings, no regressions in the blast radius.

**a3 major (snapshot publishes Cloud values): CLOSED.** `snapshot()` now reads `localTools` / `localSkills`; facts are not in the snapshot channel so no fact change was needed. Live run: snapshot before attach deep-equals snapshot after attach (tool keeps `localcanaryterm`, skill keeps `localskillterm`) while search ranks on the Cloud text (`cloudrollbackterm` -> deploy, `localcanaryterm` -> none, `cloudskillterm` -> deploy-skill). Snapshot output shape unchanged: `this.tools` was already executor-stripped and the new path strips the same `execute` / `validateInput` fields. Regression test in ratel.test.ts plus README contract update confirmed.

**a3 minor (full-corpus re-register on every apply): CLOSED.** Tool and fact applies now filter on `hasSameRetrievalDescription` vs the current effective value and skip re-registration when nothing changed; skills were already selective via replaceAll + ADR-0015. Live evidence: attach with 1 of 2 tools overridden churns only that tool; identical 200 refresh emits zero trace events; empty-overlay initial attach emits zero; override removal churns only the removed entry and restores local text; skill identical apply churns nothing. OTel adoption re-emit stays full once via `setUseCloudDefinitions(effective)` with the `:cloud` dedupe key, per CONVENTIONS.md, and repeated applies emit nothing.

**Validation (detached review worktree):** `cargo fmt --check --all`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` (389 core + all suites, 0 failed), `pnpm -r build`, `pnpm -r typecheck`, `pnpm -r lint`, `pnpm -r test` (vercel-ai-sdk 103, mastra 60): all PASS.

Still open from a3, out of this pass's scope: [minor] TS content_hash non-BMP key-order divergence (telemetry.ts:286); [nit] concurrent refresh() interleave; [nit] no test registers a tool after attach; [nit] PR body checklist stale.

No stacked-PR split recommended (unchanged from a1/a3).

:)

kayraucklnc:
Did a deep pass over the whole stack before merging. One thing here I'd fix before this ships.

**When pulling descriptions from Cloud fails halfway, we end up in a mixed state — and then tell the user we didn't.**

We apply the tool, skill and fact descriptions in parallel. If one of them fails, the others have *already* been applied. So we're left with some things using Cloud's text and some using local text, with no way back — the setup call fails, so the caller never gets a handle to retry or undo.

Then the Cloud SDK catches that failure and logs "local retrieval descriptions remain active". That message isn't true — the tools already switched over.

Reproduced it with a bad value on the skill side: the call failed, and the tool descriptions were already replaced.

Prompt to fix:

```
In src/sdk/ts/src/ratel.ts (~462-490, the definition-overrides attach path):

1. The apply is not atomic. Promise.all([tools, skills, facts]) lets each
   catalog commit its synchronous prefix before a sibling can fail, leaving a
   half-applied state. Validate and normalise the ENTIRE override set before
   touching any catalog, then either stage all three and commit only after all
   three validate, or roll applied catalogs back to their previous override map
   on failure.

2. useDefinitionOverrides is latched at ~line 465 BEFORE the applies run, so a
   FactCatalog created later inherits overrides from a failed pull. Latch it (and
   factOverrides) only after a successful apply.

3. There is no runtime validation of the remote payload at all — the types are
   compile-time only. Add a validator at the seam: reject any status that is not
   200 or 304 with a named error (today anything non-304 falls through to
   response.body.overrides, so a 500 surfaces as a raw TypeError); require body
   to be an object and overrides to be an array; require each entry to have
   kind in {tool,skill,fact} and string entryId/searchableDescription; require a
   non-empty etag on 200 (a missing etag currently means no If-None-Match is
   ever sent again, silently disabling conditional requests for the process
   lifetime).

Then in ratel-cloud-sdk src/runtime/attach.ts (~340-370): the warning says
"local Retrieval descriptions remain active" on any failure. Make it reflect
what actually happened, or make it true by fixing (1) first.
```

src/sdk/ts/src/ratel.ts
```diff
@@ -448,12 +456,50 @@ export function ratel(config: RatelConfig = {}): Ratel {
...
+  ): Promise<void> => {
+    useDefinitionOverrides = true;
+    const byKind = (kind: ExperimentalDefinitionOverride["kind"]): Map<string, string> =>
+      new Map(
+        overrides
+          .filter((override) => override.kind === kind)
+          .map((override) => [override.entryId, override.searchableDescription]),
+      );
+    factOverrides = byKind("fact");
+    await Promise.all([
```
kayraucklnc: If one of these three fails, the other two have **already** been applied. So we end up part on Cloud text and part on local text, with no way back — the call throws, so the caller never gets a handle to retry or undo.

Reproduced with a bad value on the skill side: the call failed and the tool descriptions were already replaced.

The Cloud SDK then catches this and logs *"local Retrieval descriptions remain active"*, which isn't true at that point.

One more thing right above: `useDefinitionOverrides` is set before these run, so a fact catalog created later inherits overrides from a pull that failed.

Prompt to fix:

```
In src/sdk/ts/src/ratel.ts (~462-490):

1. Make the apply atomic. Validate and normalise the ENTIRE override set before touching any
   catalog, then either stage all three and commit only once all three validate, or roll the
   applied ones back to their previous override map on failure.
2. Latch useDefinitionOverrides and factOverrides only AFTER a successful apply, not before.
3. Add a test for the partial-failure path asserting no catalog is left overridden.

Then in ratel-cloud-sdk src/runtime/attach.ts (~340-370), make the warning reflect what
actually happened rather than always claiming local descriptions are still active.
```

src/sdk/ts/src/ratel.ts
```diff
@@ -448,12 +456,50 @@ export function ratel(config: RatelConfig = {}): Ratel {
...
+    ]);
+  };
+  const experimentalAttachDefinitionOverrides = async (
+    options: ExperimentalDefinitionOverridesAttachOptions,
+  ): Promise<ExperimentalDefinitionOverridesAttachment> => {
+    let etag: string | undefined;
+    const attachment: ExperimentalDefinitionOverridesAttachment = {
+      refresh: async () => {
+        const response = await options.source.fetch(etag);
+        if (response.status === 304) return false;
```
kayraucklnc: Only 304 is handled here — everything else falls straight through to `response.body.overrides` on the next line. So a 500 or a 429 turns into a confusing crash from inside the SDK instead of a clear error.

The quieter one is the line after: a 200 with no ETag is accepted, `etag` stays undefined, and from then on we never send `If-None-Match` again. The whole conditional-request design silently becomes full pulls forever, with no error and nothing in telemetry.

What I saw feeding it bad responses:

```
body: null       -> TypeError: Cannot read properties of null
overrides: 'str' -> TypeError: overrides.filter is not a function
status: 500      -> TypeError: Cannot read properties of undefined
etag missing     -> accepted, no If-None-Match ever sent again
```

Prompt to fix:

```
Add runtime validation at the definition-overlay seam in src/sdk/ts/src/ratel.ts
(~462-490). The types there are compile-time only and nothing is checked at runtime.

1. Reject any status that is not 200 or 304 with a named DefinitionOverlayError.
2. Require body to be an object and body.overrides to be an array.
3. Require each entry to have kind in {tool,skill,fact} plus string entryId and
   searchableDescription; drop-and-warn on anything else.
4. Require a non-empty etag on a 200 response.
5. Add a per-entry length cap.

Also add an in-flight guard to refresh(): two concurrent refreshes currently both send the
same stale validator and the older response wins, silently losing the newer overlay. The
attach path already de-dupes with an #attaching promise; mirror that.
```

kayraucklnc (commented):
Inline notes on the specific lines — the half-apply item from my comment above, plus one more I hit next to it.

src/sdk/ts/src/ratel.ts
rstagi: fixed in [96871cb9](https://github.com/ratel-ai/ratel/commit/96871cb91e58e8005afc89dbf79ea6d42fad8578) and [c6253e9](https://github.com/ratel-ai/ratel-cloud-sdk/commit/c6253e96ed543843f26c2bc69e4acdda85c4be40)

src/sdk/ts/src/ratel.ts
rstagi: fixed in [96871cb9](https://github.com/ratel-ai/ratel/commit/96871cb91e58e8005afc89dbf79ea6d42fad8578)

PR #154, opened 2026-08-18
feat: add catalog protocol v2 projections

rstagi:
Stack 1/3.

Adds the experimental catalog protocol v2 projection and cross-language SDK/adapter passthrough. Public language fields are explicitly experimental (`experimentalSearchableDescription` / `experimental_searchable_description`); protocol wire fields remain neutral for interoperability. Omitting the field preserves stable retrieval behavior, including tool-schema indexing.

Stack: #154 → #155 → #152.


src/core/tests/tool_search.rs
```diff
@@ -281,12 +285,13 @@ fn search_respects_top_k_bound() {
 }
 
 #[test]
-fn search_matches_output_schema_description() {
+fn experimental_projection_does_not_match_output_schema_description() {
```
faustoq, line 352: Stable-path regression tests were converted into experimental-path tests, not added alongside. 

1. `search_matches_output_schema_description` -> `experimental_projection_does_not_match_output_schema_description`

2. `search_matches_nested_object_description` -> `experimental_projection_does_not_match_nested_object_description`

3. `search_matches_array_items_description` -> `experimental_projection_does_not_match_array_items_description`

4. `search_matches_enum_value` -> `experimental_projection_does_not_match_enum_value`

5. `search_matches_input_param_description` -> `experimental_projection_does_not_match_input_param_description`

6. `search_matches_input_param_name` -> `experimental_projection_does_not_match_input_param_name`


ADR-0021's central promise is "**_Existing callers retain schema-inclusive tool ranking_**".
After this PR, nothing tests it. Nested-object, array-items, enum, param-name indexing on the stable path all have zero coverage. I think these should be duplicated (not replaced).

src/core/src/indexing.rs
```diff
@@ -103,30 +110,10 @@ mod tests {
...
-        let encoding_idx = text.find("encoding").expect("encoding token missing");
-        assert!(
-            path_idx < encoding_idx,
-            "expected schema-defined order (path before encoding) in: {text}"
-        );
-    }
-
-    #[test]
-    fn searchable_text_omits_json_structure_keywords() {
+    fn stable_searchable_text_keeps_schemas() {
```
faustoq, line 120: Same as above, can we add the new test along side the existing one, instead of replacing it?

src/core/tests/tool_search.rs
```diff
@@ -248,14 +257,8 @@ fn search_ranks_stronger_match_above_weaker() {
...
 
-    assert!(
-        hits.len() >= 2,
-        "expected both tools to match, got {}",
-        hits.len()
-    );
+    assert_eq!(hits.len(), 1);
     assert_eq!(hits[0].tool_id, "strong");
-    assert_eq!(hits[1].tool_id, "weak");
-    assert!(hits[0].score > hits[1].score);
```
faustoq, line 258: this was `search_ranks_stronger_match_above_weaker`
it tested BM25 relative ordering, not this feature, now the `hits[0].score > hits[1].score` assertion is gone. can we restore the original test alongside this one?

src/core/src/indexing.rs
```diff
@@ -103,30 +110,10 @@ mod tests {
...
-            !text.contains("\"type\""),
-            "raw type quoting leaked: {text}"
-        );
-        assert!(
-            !text.contains("\"properties\""),
-            "properties leaked: {text}"
-        );
-        assert!(!text.contains('{'), "JSON braces leaked: {text}");
+        assert!(text.contains("path"), "input schema missing: {text}");
+        assert!(text.contains("encoding"), "output schema missing: {text}");
```
faustoq: `encoding` is a property of `input_schema` (line 93), not of `output_schema`, which is `json!({})` in the fixture. 

The message is wrong, and the test doesn't cover the **output** schema: on the stable path, `output_schema` indexing is left untested.

faustoq (commented):
The feature itself looks solid. Clean design, consistent across all four layers (core, both native bindings, both SDKs, both adapters), and the ADR housekeeping is thorough. 

All my comments are the same theme: the new tests were added by converting the stable-path ones rather than sitting beside them. ADR-0021 promises that existing callers retain schema-inclusive tool ranking, and after this PR nothing verifies that. Duplicating instead of replacing should be a small change and it puts the regression net back.

src/core/tests/tool_search.rs
rstagi: fixed in [8bab19e5](https://github.com/ratel-ai/ratel/commit/8bab19e5d15d19a73dd43d4f54ed2ae4c94faf51)

src/core/src/indexing.rs
rstagi: fixed in [f521693a](https://github.com/ratel-ai/ratel/commit/f521693a353badc0ffb6b267bd500403625dc859)

src/core/tests/tool_search.rs
rstagi: fixed in [1f766559](https://github.com/ratel-ai/ratel/commit/1f7665598e43f5337016ebbad2ce694cd1823ef2)

src/core/src/indexing.rs
rstagi: fixed in [3e56ae80](https://github.com/ratel-ai/ratel/commit/3e56ae80b97d58c8bbcb438ca613d79016a2c3b6)

faustoq (approved):
LGTM!

PR #155, opened 2026-08-18
feat: emit catalog definition telemetry

rstagi:
Stack 2/3; depends on #154.

Adds experimental cross-language catalog-definition runtime/OTel telemetry, RFC 8785 canonicalization, shared conformance vectors, and contract documentation.

Definition production is disabled by default. Runtime events require `experimentalCatalogDefinitions: true` / `experimental_catalog_definitions=True`; OTel additionally requires `RATEL_EXPERIMENTAL_CATALOG_DEFINITIONS=true` plus an EventRecord content-capture mode. Wire event and attribute names remain neutral.

Stack: #154 → #155 → #152.


kayraucklnc:
Did a deep pass over the whole stack before merging. Three things here worth fixing.

---

**1. When a payload is too big, we throw away the wrong parts first.**

If a definition is too large to send, we start dropping fields until it fits. But the order we drop them in means the *name, the ID and the kind go first* — and the giant schema that caused the problem stays. So Cloud receives a definition with no idea which tool it belongs to, or even whether it's a tool at all.

Checked what actually arrives — no `kind`, no `id`, no `name`, no `description`. Which fields survive depends on the order they happen to be listed in, not on what matters.

**2. A couple of big tools can knock out unrelated telemetry.**

There's no size limit at all on the other sending path. A 2 MB schema goes out as a single 2-million-character value. A few of those together exceed the receiving limit and the *whole batch* is rejected — quietly dropping traces and logs that had nothing to do with catalogs.

**3. The three languages disagree on very large numbers.**

If a schema contains a number bigger than roughly 9 quadrillion, Python skips sending the definition entirely, while Rust and TypeScript send it with the number silently rounded. Two side effects: the same catalog produces different results depending on which SDK you use, and once rounded, *editing that number never registers as a change again* — Cloud stops seeing updates to that tool permanently.

Worth picking one behaviour and pinning it with a shared test. My vote is Python's (skip and move on) since rounding produces a value we can't verify later, but either is fine as long as all three match.

---

Prompt to fix:

```
Three fixes in the ratel catalog-definition telemetry path.

1. src/sdk/ts/src/runtime-events.ts (~435-460, isProductFactField) and the
   Python twin in ratel_ai/runtime_events.py (_is_product_fact_field):
   the definition vocabulary was never added, so kind/id/name/description/tags
   are the first fields dropped by the size-bounded fallback. Add them to the
   preserved set (kind, id, name and content_hash at minimum) and change the
   drop order so input_schema and output_schema go FIRST, not last.

2. src/sdk/ts/src/telemetry.ts (~252-270) and ratel_ai/telemetry.py (~184-196):
   there is no size cap on the OTel lane. Add one for the schema attributes;
   above the cap, skip the schema and set a flag on the record rather than
   emitting a multi-megabyte attribute that fails the whole batch export.

3. Pick one canonical behaviour for integers outside +/-2^53. Today Python
   skips the event (rfc8785 IntegerDomainError) while Rust
   (src/core/src/trace/event.rs:647) and TS round via
   serde_json_canonicalizer / canonicalize. Make all three agree, add a
   conformance vector in src/telemetry/conformance/fixtures.json covering it,
   and correct CONVENTIONS.md — it currently documents NaN/infinity rejection,
   which is unreachable, and says nothing about the case that actually happens.

Also related: while fixing (3), note that content_hash is currently computed
over the rounded value while the payload carries the exact one, so Cloud can
never re-derive the hash from what it received.
```

kayraucklnc (commented):
Inline notes on the specific lines — same three items as my comment above, anchored to the code so they're easy to act on.

src/sdk/ts/src/runtime-events.ts
```diff
@@ -15,6 +15,7 @@ const REQUIRED_ENVELOPE_FIELDS = new Set([
 
 /** Frozen remotely publishable v1 event names from ADR-0020. */
 export const RUNTIME_EVENT_TYPES = [
+  "catalog_definition",
```
kayraucklnc, line 28: Heads up on adding this to the lane: events here go through the size-trimming logic, and that logic doesn't know about the definition fields. So when a payload is too big, `kind`, `id` and `name` are the **first** things dropped — and the giant schema that caused the problem stays.

Cloud ends up with a definition it can't identify. This is what actually arrives:

```
["v","event_id","ts","session_id","source_id","type",
 "output_schema","searchable_description",
 "searchable_description_overridden","content_hash"]
```

Which fields survive depends on the order they happen to be listed in, not on what matters.

Prompt to fix:

```
In src/sdk/ts/src/runtime-events.ts (isProductFactField, ~435-460) and the Python
twin _is_product_fact_field in ratel_ai/runtime_events.py: the definition vocabulary was
never added, so kind/id/name/description/tags are the first fields the size-bounded
fallback deletes.

1. Add them to the preserved set — kind, id, name and content_hash at minimum.
2. Reverse the drop order so input_schema and output_schema go FIRST, not last.
3. Add a test with an oversized schema asserting kind/id/name survive the trim.
```

src/sdk/ts/src/telemetry.ts
```diff
@@ -181,6 +195,92 @@ export interface RuntimeEventProjection {
...
+  const contentHash = createHash("sha256").update(canonicalJson(content), "utf8").digest("hex");
+  return {
+    [RATEL_CATALOG_KIND]: kind,
+    [RATEL_CATALOG_ID]: definition.id,
+    [RATEL_CATALOG_NAME]: definition.name,
+    [RATEL_CATALOG_DESCRIPTION]: definition.description,
+    [RATEL_CATALOG_TAGS]: tags,
+    ...(kind === "tool"
+      ? {
+          [RATEL_CATALOG_INPUT_SCHEMA]: canonicalJson(inputSchema),
```
kayraucklnc: There's no size limit on this. A tool with a 2 MB schema becomes a single 2-million-character value — I measured 2,000,039 characters.

A few of those going out together exceed the receiving limit, and that rejects the **whole batch** — silently dropping traces and logs that had nothing to do with catalogs. So one big tool takes out unrelated telemetry.

Prompt to fix:

```
In src/sdk/ts/src/telemetry.ts (~252-270) and ratel_ai/telemetry.py (~184-196):
the catalog schema attributes have no size cap.

1. Add a per-attribute cap for input_schema/output_schema.
2. Above the cap, omit the schema and set a truncation flag on the record rather than
   emitting a multi-megabyte attribute that fails the entire log batch export.
3. Test that an oversized schema still produces a usable record and does not fail the
   batch.
```

src/core/src/trace/event.rs
```diff
@@ -504,6 +557,99 @@ pub enum TraceEvent {
...
+            output_schema: output_schema.map(Box::new),
+            searchable_description: searchable_description.into(),
+            searchable_description_overridden,
+            content_hash,
+        }
+    }
+}
+
+fn catalog_definition_hash(content: &CatalogDefinitionContent<'_>) -> String {
+    let canonical = serde_json_canonicalizer::to_vec(content)
```
kayraucklnc: This rounds very large numbers rather than rejecting them, which has a nastier effect than it looks. Measured directly:

```
18446744073709551615 -> "18446744073709552000"
9007199254740993     -> "9007199254740992"
```

Two consequences. Different schemas produce the same hash — I registered four genuinely different schemas and only two events came out. And once a number is in that range, **editing it never registers as a change again**, so Cloud stops seeing updates to that tool permanently.

Also, the payload carries the exact original number while the hash is over the rounded one, so Cloud can never re-derive or verify `content_hash` from what it received.

And the three SDKs disagree here: **Python skips the event entirely, Rust and TS round and send.** Which one do we want as canonical? My vote is Python's, since a rounded value can't be verified later — but either is fine as long as all three match and a shared test pins it.

Prompt to fix:

```
Pick one canonical behaviour for integers outside +/-2^53 in the catalog-definition
path, then make all three languages agree:
  - Rust: src/core/src/trace/event.rs:647-651 (serde_json_canonicalizer, currently rounds)
  - TS:   src/sdk/ts/src/telemetry.ts (canonicalize, currently rounds)
  - Py:   ratel_ai/telemetry.py:176 (rfc8785 IntegerDomainError, currently skips)

Add a conformance vector in src/telemetry/conformance/fixtures.json covering it, asserted
by all three suites. Also correct CONVENTIONS.md — it documents NaN/infinity rejection,
which is unreachable through serde_json::Value, and says nothing about this case, which is
the one that actually happens.

While here: the .expect() on line 649 is a panic on a registration hot path behind an FFI
boundary. Make it fail-open instead.
```

rstagi: fixed in [bb583c0a](https://github.com/ratel-ai/ratel/commit/bb583c0ab171f2e3e43c09dbd7ff79bc2c74cdb9)

src/sdk/ts/src/telemetry.ts
rstagi: fixed in [aa2e7a42](https://github.com/ratel-ai/ratel/commit/aa2e7a427b5a7523fec615e2e58abffe69e53009)

src/sdk/ts/src/runtime-events.ts
rstagi: fixed in [98b43315](https://github.com/ratel-ai/ratel/commit/98b43315f71a6f46f304bea74550934994f4c28b)

PR #159, opened 2026-08-31
feat: adaptive ranking and retrieval scoring — cluster admission, negative feedback, and a readable score

bercaakbayir:
feat: adaptive ranking and retrieval scoring

Fixes the misranking reported in the 2026-08-11 Kestral investigation: read-phrased queries served write operations, and the score shown next to them couldn't be read as confidence because it wasn't one.

Three groups — how clusters form, what an edge is worth, and what a score means.

Clustering

Admission by member coverage, not centroid proximity alone. Prefilter cos(query, centroid) ≥ τ, count how many of the cluster's members the query also clears τ against, and admit only on count ≥ min(members, max(2, ceil(coverage × members))). Matching one member is single-link chaining — A joins because of B, and the cluster grows into whatever B happened to bridge to. Matching an average is worse, because the average of two intents resembles neither. The dense tier had no per-member guard while the lexical tier always had one.

ClusterPolicy { similarity, coverage } exposed per catalog. τ is model-dependent (a cosine of 0.70 means different things on different embedding models) and corpus-dependent (a narrow catalog and a broad one want different granularity), so no constant can be right. Out of range is rejected rather than clamped, and the policy is recorded on the graph it clustered — two producers at different settings would otherwise disagree about what a cluster means while both claiming v: 1. A mismatch reports PolicyDrift, not Paused: the clusters are still coherent, just coarser or finer, and rebuilding can't revisit boundaries.

Centroid running-mean fix. absorb_vector rescaled an already-normalised centroid by the fold count, which asserted the accumulated history had full length — true only when every member points the same way. The mean drifted instead of converging, over-weighting early members. Now a true incremental mean, with cohesion and vector_n persisted so a reloaded graph keeps folding correctly rather than treating its centroid as a single sample.

Edge weighting

Inverse cluster frequency. An edge weight is a count of confirmed invocations, but serving that raw lets a capability invoked across many intents rank on volume rather than on answering this question — the mechanism behind the reported failure. Each edge is scaled by 1 + ln(clusters / clusters naming it), counted over every cluster's raw edges so the order is a property of the graph, not of which catalog is attached. Where counts previously tied, order fell through to the id tie-break.

Impressions as a denominator. We recorded invocations and never what was shown, so a capability surfaced twelve times and taken once was indistinguishable from one surfaced once and taken once. Edges are now damped by min(1, (invoked + 3) / (considered + 3)), for tools and skills alike. Counted at confirm, so an abandoned search teaches nothing, and once per search window. Only ids ranked at or above the invoked one count — one listed below the one the caller took was very likely never read, and counting it as refused would penalise it for our own ranking. Neutral where no impressions exist (3/3 = 1), and clamped so it can only ever cost an edge, never promote one.

Scoring

Hybrid fuses on normalised scores, not rank positions. (1-w)·bm25/Σidf + w·cos, plus the usage arm at the share it already had. ADR-0011 rejected this because BM25 is unbounded and cosine is [-1, 1]; both arms now carry an absolute [0, 1] value, so they share a scale for the first time. RRF's reachable maximum was "first in every arm", which says nothing about match quality — a query the catalog answers precisely and one it cannot answer at all both returned ~1.0 at the top. Ordering is decided before the clamp, so a candidate the usage arm promotes past the content ceiling keeps its position.

DenseWeight — the dense/lexical split is a per-catalog setting rather than a constant, because the balance is a property of the catalog. A catalog of natural-language descriptions gives the dense arm most of the signal; one keyed on exact identifiers, error codes, or internal jargon gives BM25 purchase the first has none of. 0 is pure lexical, 1 pure dense, both reachable.

relevance on SearchHit / SkillHit. score sits on three incomparable scales, so it was never displayable, and normalising it yourself is how a rank position gets read as certainty. Each method carries the rule its own scale admits: (cos + 1) / 2 for cosine, score / Σ idf(query terms) for raw BM25 — the share of the query's discriminating mass a hit captured, which also exposes when the catalog has no vocabulary for the question — and for hybrid the fused score itself, already absolute. Not a calibrated confidence: nothing was fitted to whether a hit was the one the caller went on to invoke.

BM25 b raised 0.4 → 0.75. The 0.4 was chosen when a tool's document was its description plus every schema token, where length mostly reflected how many arguments a tool took.

Protocol

protocol/v1 gains cluster_policy, cohesion, vector_n, surfaced_tools and surfaced_skills — all optional, omitted when empty, GRAPH_VERSION unchanged, so a graph written before them stays byte-identical. Impressions are documented as a denominator a consumer may damp with, never an edge: an id present in an impression map but absent from the matching edge map has no edge and must not be promoted.


rstagi (commented):
some major and some minor things, overall the implementation is solid but it needs some changes before merging :) see dedicated comments below 👇

src/core/src/usage.rs
```diff
@@ -554,46 +783,187 @@ impl Intent {
...
+                *m += (v - *m) / n;
             }
-            // A first vector — or one of a different width (the embedding model
-            // changed) — starts the mean fresh rather than blending across spaces.
-            _ => {
-                self.vector_n = 0;
-                vector.to_vec()
+        }
+        let mean = self.mean.clone().expect("set on both arms above");
+        self.cohesion = norm(&mean);
```
rstagi, line 890: [blocker] generated cohesion can make the graph immediately unreadable

Recomputing the norm of an f32-normalized embedding can produce slightly more than `1.0`. This value is serialized, then `IntentGraph::from_json` rejects it.

Reproduced through the shipped TypeScript example: turn 7 fails with `cohesion 1.0000002 outside (0, 1]`.

Clamp producer-side cohesion to `1.0` here and in `rebuild_centroids`, then add a non-axis-aligned serialize/load regression test.

src/core/src/usage.rs
```diff
@@ -518,20 +692,75 @@ pub struct Intent {
...
-    #[serde(skip)]
+    /// inflate the weight.
+    ///
+    /// Carried on the wire so learning resumes at the right weight after a
+    /// round-trip. Without it the running mean restarted at one sample, and the
+    /// next single observation could drag a fifty-member cluster's centroid
+    /// halfway across — every save/load cycle made centroids plastic again.
+    /// Absent on the wire means one, the weight the pre-accumulator code gave a
+    /// reloaded centroid.
+    #[serde(default = "one_u32", skip_serializing_if = "is_one_u32")]
```
rstagi, line 703: [major] lexical graphs violate the published vector_n schema

Lexical clusters retain `vector_n = 0`, but this predicate serializes every value except `1`. The resulting graph contains `"vector_n":0`, while the v1 schema requires a minimum of `1`.

Ratel's own loader currently accepts this and ranking ignores the field, so this is not an immediate runtime blocker. It is still a producer/schema contract violation on the default BM25 path.

Omit values `<= 1` and add a lexical producer conformance test.

src/core/src/usage_learner.rs
```diff
@@ -214,16 +288,42 @@ pub(crate) fn replay_log_into(
...
                 // Re-arming with the same text is idempotent: a capability
-                // search fans one question to both catalogs, and both of those
-                // land before any invoke, so the turn still credits once.
-                pending.insert(session, (query, false));
+                // search fans one question to both catalogs, and both land
+                // before any invoke, so the turn still credits once. Each fills
+                // its OWN slot — one slot would let the skill search's ids
+                // overwrite the tool search's and land in the wrong map.
+                let entry = pending.entry(session).or_default();
+                if entry.query != query {
```
rstagi, line 317: [major] replay fails to re-arm repeated same-text searches

Replay resets `credited` only when the query text changes. Live learning calls `arm_credit` for every search. Consequently `search q -> invoke -> search q -> invoke` records support `1` during replay and `2` live, despite identical events.

This directly weakens the seeded usage arm and violates the documented replay/live parity. Reset credit for each new search while preserving tool/skill fan-out, and assert complete intent parity in the existing replay test.

src/core/src/search.rs
```diff
@@ -35,21 +140,67 @@ impl Bm25Index {
...
+    /// Not a hard maximum: a short document repeating a term can exceed it, so
+    /// callers clamp.
+    pub(crate) fn query_ceiling(&self, query: &str) -> f32 {
+        let n = self.doc_count as f32;
+        if n == 0.0 {
+            return 0.0;
         }
+        let mut terms = tokenizer().tokenize(query);
+        terms.sort_unstable();
+        terms.dedup();
```
rstagi, line 196: [major] BM25 scoring and its normalization disagree about repeated terms

The engine scores every query-token occurrence, while `query_ceiling` deduplicates them. A repeated term can therefore double the numerator without changing the denominator, saturating relevance and potentially changing hybrid ordering.

Make multiplicity consistent on both sides, either by deduplicating before engine scoring or counting occurrences in the ceiling. Add a duplicate-term regression test.

src/core/src/tool_registry.rs
```diff
@@ -320,7 +454,7 @@ impl ToolRegistry {
         for (id, cluster_members) in &members {
             let (vectors, fp) = self
                 .dense
-                .embed_texts_with_identity(cluster_members, self.sink.as_ref())?;
+                .embed_queries_with_identity(cluster_members, self.sink.as_ref())?;
```
rstagi, line 457: [major] the query-side embedding correction is missing from graph seeding

Rebuilding correctly embeds intent members through `embed_queries_with_identity`, but `build_intent_graph` still sends those same past queries through `embed_texts_with_identity`.

This contradicts the documented claim that replay clusters in the same embedding space as live learning and can fail for configured asymmetric embedders. Use the query-side batch API in the seeding path and test it with an embedder whose query and document vectors differ.

protocol/v1/conformance/vectors.json
```diff
@@ -488,8 +488,12 @@
...
+      "`vector_n` is provenance for learning only \u2014 it weights the next fold into `centroid`. Retrieval results MUST be identical whatever it contains.",
+      "`seeded_support` is provenance only — it records how many of `support` came from a seeding pass rather than live traffic. Retrieval results MUST be identical whatever it contains.",
+      "`surfaced_tools` and `surfaced_skills` are provenance only — it counts how many of a cluster's searches put each capability in front of the caller, so a consumer can tell a capability that is shown constantly and rarely used from one shown rarely and always used. A consumer MUST NOT let it affect ranking, and MUST NOT treat an id present in an impression map but absent from the matching edge map as an edge: what retrieval returned is not evidence, only what was invoked is. The two maps exist because tool and skill ids are different id spaces: an id in one MUST NOT be read against the other."
```
rstagi, line 496: [minor] this consumer rule contradicts the schema and implementation

This says impression maps must not affect ranking. The schema and protocol README say consumers may use them to damp ranking, and `Intent::ranked` does so.

Update this rule to match the schema. Preserve the separate invariant that an impression-only id must never become an edge.

src/core/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "ratel-ai-core"
-version = "0.11.0" # core-v* release unit (ADR-0008)
+version = "0.12.0-rc.4" # core-v* release unit (ADR-0008)
```
rstagi, line 3: [minor] bump versions before publishing the next candidate

The rc.4 versions already identify artifacts published from `623ae13d`, while HEAD contains later behavioral changes. Registries cannot republish those versions.

This is a release-time requirement rather than a current runtime defect. Bump core and SDK manifests to rc.5 before the next candidate is tagged.

src/core/src/usage.rs
bercaakbayir: (421c4a1) cohesion is computed by taking the length (norm) of an averaged vector, and floating-point rounding could occasionally push that length a tiny bit past 1.0. Our own file loader rejects any cohesion value outside 0–1, so we'd sometimes write a file we couldn't read back. I fixed it by clamping the result to at most 1.0 right where it's calculated, in both functions that compute it (absorb_vector and rebuild_centroids), so the value can never leave that valid range.

src/core/src/usage.rs
bercaakbayir: (421c4a1) Clusters that were only ever matched by keyword (never given a real vector) internally store vector_n = 0. Our rule for what to leave out of the saved file only skipped this field when it was exactly 1, so a 0 was still being written, but the file format's rules say this field can't be 0 if it's present at all. I fixed it by changing that skip rule to cover both 0 and 1, since either value just means "nothing meaningful to report."

src/core/src/usage_learner.rs
bercaakbayir: (234cd4d) entry.credited = false; was added inside the loop that processes each search event during replay, placed so it runs every time a search happens, right before the code moves on to record what that search surfaced. Previously, that reset only happened as a side effect of the whole tracking entry being replaced when the query text changed. By being pulled out into its own unconditional line, it now runs on every search, matching how the live-traffic code already re-arms credit on every search regardless of repeated text.

src/core/src/search.rs
bercaakbayir: (a00073c) The word-deduplication logic that existed in one function (query_ceiling)  was pulled out into its own small shared function, distinct_terms. The scoring function (search) was then made to call that same shared function before the query is handed off to the scoring engine, with the now-unique words joined back into a string. Because both functions now go through the exact same deduplication step, there's no longer any way for them to disagree about how many times a repeated word should count.

src/core/src/tool_registry.rs
bercaakbayir: (91aea85) A one-word swap was made in the function call: `embed_texts_with_identity` became `embed_queries_with_identity`. Both functions already existed in the codebase and were correctly implemented for their respective purpose, one for documents, one for queries, the seeding code was just calling the wrong one. `embed_texts_with_identity` was also removed afterward since that was its only caller and it would've been left as unused code.

protocol/v1/conformance/vectors.json
bercaakbayir: (c62bec6) The contradicting sentence was edited to match the schema's own wording:
"MUST NOT let it affect ranking" became "MAY let it damp ranking."

src/core/Cargo.toml
bercaakbayir: (6f5cd15) The version string was bumped from rc.4 to rc.5 in every relevant file:
- core: 0.12.0-rc.4 → 0.12.0-rc.5
- ts: 0.13.0-rc.4 → 0.13.0-rc.5
- py: 0.13.0rc4 → 0.13.0rc5

rstagi (approved):

PR #160, opened 2026-08-31
fix(sdk): report a failed tool call on both telemetry channels

faustoq:
## Problem

MCP reports a failed tool call in the result (`isError: true`), not by raising an error. 

Both telemetry channels treated only a thrown exception as a failure:

- `traceExecuteTool` set the `execute_tool` span to `OK` for any return that did not throw
- `runToolInvocation` emitted `invoke_end` for those same returns

So a failed call and a successful one were recorded identically, and the dashboard's error rate could not report tool failures at all.

## Solution

A shared `reportsFailure` predicate, exported `@internal` from the telemetry module so both channels agree on what counts as a failure. The span now closes as `ERROR`, and the runtime event becomes `invoke_error` through the existing `reject` path, so no new event type is introduced and nothing downstream needs updating.

Applied to both the TypeScript and Python SDKs.

The failure text is generic (`the tool reported a failure`) because result content sits behind the content-capture gate.

## Notes

The other four unconditional `set_status(...OK)` calls in `ratel_ai/telemetry.py` were checked and do not need this: `trace_search`, `trace_search_async`, `trace_skill_load` and `trace_upstream_register` do not execute tools, so the `isError` convention never reaches them.

### DIFF
--- a/src/sdk/python/ratel_ai/catalog.py
+++ b/src/sdk/python/ratel_ai/catalog.py
@@ -32,6 +32,7 @@ from .telemetry import (
     SEARCH_TARGET_TOOL,
     RuntimeEventProjection,
     record_catalog_definitions,
+    reports_failure,
     trace_execute_tool,
     trace_search,
     trace_search_async,
@@ -1336,14 +1337,21 @@ class ToolCatalog:
                     result = await result
                 terminal_projection = projection.copy()
                 terminal_projection["event_id"] = new_runtime_event_id()
-                self._registry.record_event(
+                terminal: dict[str, Any] = (
                     {
+                        "type": "invoke_error",
+                        "tool_id": tool_id,
+                        "took_ms": _elapsed_ms(started),
+                        "error": "the tool reported a failure",
+                    }
+                    if reports_failure(result)
+                    else {
                         "type": "invoke_end",
                         "tool_id": tool_id,
                         "took_ms": _elapsed_ms(started),
-                    },
-                    terminal_projection,
+                    }
                 )
+                self._registry.record_event(terminal, terminal_projection)
                 return result
             except asyncio.CancelledError as err:
                 terminal_projection = projection.copy()
--- a/src/sdk/python/ratel_ai/telemetry.py
+++ b/src/sdk/python/ratel_ai/telemetry.py
@@ -355,6 +355,11 @@ def _normalize_content(value: Any) -> Any:
     return str(value)
 
 
+def reports_failure(result: Any) -> bool:
+    """MCP reports a failed call in the result (`isError: True`), not by raising."""
+    return isinstance(result, dict) and result.get("isError") is True
+
+
 async def trace_execute_tool(
     tool_id: str,
     args: dict[str, Any],
@@ -388,7 +393,10 @@ async def trace_execute_tool(
             span.set_attribute(GEN_AI_TOOL_CALL_RESULT, _safe_json(result))
         if _capture_content_on_event():
             _add_tool_content_event(tool_id, args, projection["event_id"], result=result)
-        span.set_status(Status(StatusCode.OK))
+        if reports_failure(result):
+            span.set_status(Status(StatusCode.ERROR, "the tool reported a failure"))
+        else:
+            span.set_status(Status(StatusCode.OK))
         return result
 
 
--- a/src/sdk/python/ratel_ai/telemetry.py
+++ b/src/sdk/python/ratel_ai/telemetry.py
@@ -356,8 +356,14 @@ def _normalize_content(value: Any) -> Any:
 
 
 def reports_failure(result: Any) -> bool:
-    """MCP reports a failed call in the result (`isError: True`), not by raising."""
-    return isinstance(result, dict) and result.get("isError") is True
+    """MCP reports a failed call in the result (`isError: True`), not by raising.
+
+    The client returns a `CallToolResult` model, not a dict, and 2.x renamed the
+    field to `is_error` (alias `isError`), hence both shapes and both spellings.
+    """
+    if isinstance(result, dict):
+        return result.get("isError") is True or result.get("is_error") is True
+    return getattr(result, "is_error", None) is True or getattr(result, "isError", None) is True
 
 
 async def trace_execute_tool(
--- a/src/sdk/python/tests/test_catalog.py
+++ b/src/sdk/python/tests/test_catalog.py
@@ -673,6 +673,19 @@ async def test_invoke_emits_start_then_end_telemetry() -> None:
     assert "took_ms" in events[1]
 
 
+async def test_invoke_emits_invoke_error_when_the_tool_reports_a_failure() -> None:
+    catalog = ToolCatalog(trace=TraceSinkConfig(kind="memory", session_id="s"))
+    await catalog.register(
+        _read_file_tool(
+            lambda args: {"isError": True, "content": [{"type": "text", "text": "boom"}]}
+        )
+    )
+    catalog.drain_trace_events()
+    await catalog.invoke("read_file", {"path": "/a"})
+    types = [e["type"] for e in catalog.drain_trace_events()]
+    assert types == ["invoke_start", "invoke_error"]
+
+
 async def test_invoke_emits_error_telemetry_and_reraises() -> None:
     def boom(args):
         raise RuntimeError("kaboom")
--- a/src/sdk/python/tests/test_telemetry.py
+++ b/src/sdk/python/tests/test_telemetry.py
@@ -118,6 +118,27 @@ async def test_execute_tool_span_attributes(exporter: Any) -> None:
 
 
 @pytest.mark.asyncio
+async def test_execute_tool_span_is_error_when_the_tool_reports_a_failure(
+    exporter: Any,
+) -> None:
+    catalog = ToolCatalog()
+    await catalog.register(
+        ExecutableTool(
+            id="failing_tool",
+            name="failing_tool",
+            description="Reports a failure in its result instead of raising.",
+            input_schema={"properties": {}},
+            output_schema={"properties": {}},
+            execute=lambda args: {"isError": True, "content": [{"type": "text", "text": "boom"}]},
+        )
+    )
+    await catalog.invoke("failing_tool", {})
+
+    spans = _spans_named(exporter, "execute_tool failing_tool")
+    assert spans[0].status.status_code == StatusCode.ERROR
+
+
+@pytest.mark.asyncio
 async def test_search_span_shares_runtime_event_id(exporter: Any) -> None:
     catalog = ToolCatalog()
     skills = SkillCatalog()
--- a/src/sdk/python/tests/test_telemetry.py
+++ b/src/sdk/python/tests/test_telemetry.py
@@ -139,6 +139,33 @@ async def test_execute_tool_span_is_error_when_the_tool_reports_a_failure(
 
 
 @pytest.mark.asyncio
+async def test_execute_tool_span_is_error_on_the_mcp_result_model(exporter: Any) -> None:
+    """The span channel reads the failure off a `CallToolResult`, not only off a dict."""
+    pytest.importorskip("mcp", reason="install ratel-ai[mcp] to run MCP result-shape tests")
+    from mcp import types
+
+    failure = types.CallToolResult(
+        content=[types.TextContent(type="text", text="boom")],
+        isError=True,
+    )
+    catalog = ToolCatalog()
+    await catalog.register(
+        ExecutableTool(
+            id="failing_model_tool",
+            name="failing_model_tool",
+            description="Returns the result model an mcp client hands back on a failed call.",
+            input_schema={"properties": {}},
+            output_schema={"properties": {}},
+            execute=lambda args: failure,
+        )
+    )
+    await catalog.invoke("failing_model_tool", {})
+
+    spans = _spans_named(exporter, "execute_tool failing_model_tool")
+    assert spans[0].status.status_code == StatusCode.ERROR
+
+
+@pytest.mark.asyncio
 async def test_search_span_shares_runtime_event_id(exporter: Any) -> None:
     catalog = ToolCatalog()
     skills = SkillCatalog()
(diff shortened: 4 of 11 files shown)


src/sdk/python/ratel_ai/telemetry.py
```diff
@@ -355,6 +355,11 @@ def _normalize_content(value: Any) -> Any:
     return str(value)
 
 
+def reports_failure(result: Any) -> bool:
+    """MCP reports a failed call in the result (`isError: True`), not by raising."""
+    return isinstance(result, dict) and result.get("isError") is True
```
kayraucklnc: **Blocking — this is `False` for every real MCP result.**

`mcp.py:231` wires each executor to `await session.call_tool(...)`, and the mcp Python client returns a pydantic **`CallToolResult`**, never a `dict`. So `isinstance(result, dict)` fails on every upstream call, successful or not, and this predicate can only ever be true for a hand-built dict.

Verified against a real in-memory MCP server on this branch: `reports_failure(result) -> False`, events `['invoke_start', 'upstream_invoke', 'invoke_end']`.

Also note `mcp` 2.x renamed the model field to `is_error` (keeping `isError` only as a serialization alias), so an attribute check on `isError` alone would fix 1.x and still miss the version CI resolves today.

Verified fix (full suite green, ruff + mypy clean, correct on 1.29.1 and 2.1.1):

```python
if isinstance(result, dict):
    return result.get("isError") is True or result.get("is_error") is True
return getattr(result, "is_error", None) is True or getattr(result, "isError", None) is True
```

The dict branch still matters — ratel's own tools (`capabilities.py`, `skill_tools.py`) do return dicts.

src/sdk/python/ratel_ai/catalog.py
```diff
@@ -1336,14 +1337,21 @@ async def _run(projection: RuntimeEventProjection) -> Any:
...
                 terminal_projection["event_id"] = new_runtime_event_id()
-                self._registry.record_event(
+                terminal: dict[str, Any] = (
                     {
+                        "type": "invoke_error",
+                        "tool_id": tool_id,
+                        "took_ms": _elapsed_ms(started),
+                        "error": "the tool reported a failure",
+                    }
+                    if reports_failure(result)
```
kayraucklnc, line 1347: Consequence of the predicate bug above: for MCP-proxied tools this condition is **never** true, so the `invoke_error` branch is unreachable and every upstream failure falls through to `invoke_end`.

The structure here is fine — it's the predicate that needs fixing, not this. Flagging it so the dead branch is visible.

src/sdk/python/tests/test_catalog.py
```diff
@@ -673,6 +673,19 @@ async def test_invoke_emits_start_then_end_telemetry() -> None:
     assert "took_ms" in events[1]
 
 
+async def test_invoke_emits_invoke_error_when_the_tool_reports_a_failure() -> None:
+    catalog = ToolCatalog(trace=TraceSinkConfig(kind="memory", session_id="s"))
+    await catalog.register(
+        _read_file_tool(
+            lambda args: {"isError": True, "content": [{"type": "text", "text": "boom"}]}
```
kayraucklnc, line 680: **Blocking (test gap).** This returns a literal `dict` — the *wire* shape. A real MCP call never produces one; it produces a `CallToolResult` model. So this test passes while the actual integration is broken.

It isn't a vacuous test: I reverted the `catalog.py` fix and it went red, so it is genuinely bound to the implementation. It's bound to the wrong *input*, which is much harder to spot.

Please keep this one, but add a real-client regression test in `tests/test_mcp.py` using the existing `_memory_client_session` harness. That one would have failed on this branch.

src/sdk/python/tests/test_telemetry.py
```diff
@@ -117,6 +117,27 @@ async def test_execute_tool_span_attributes(exporter: Any) -> None:
...
+) -> None:
+    catalog = ToolCatalog()
+    await catalog.register(
+        ExecutableTool(
+            id="failing_tool",
+            name="failing_tool",
+            description="Reports a failure in its result instead of raising.",
+            input_schema={"properties": {}},
+            output_schema={"properties": {}},
+            execute=lambda args: {"isError": True, "content": [{"type": "text", "text": "boom"}]},
```
kayraucklnc, line 132: Same test gap as `test_catalog.py`: `execute` returns a literal dict, so the span assertion passes against an input real MCP never yields.

Worth covering the span channel from the real client path too, so both channels are proven on a `CallToolResult` rather than on a dict.

src/sdk/ts/src/telemetry.ts
```diff
@@ -734,6 +734,18 @@ export function upstreamFromToolId(toolId: string): string | undefined {
...
 
+/**
+ * MCP reports a failed call in the result (`isError: true`), not by throwing.
+ * @internal
+ */
+export function reportsFailure(result: unknown): boolean {
+  return (
+    typeof result === "object" &&
+    result !== null &&
+    (result as { isError?: unknown }).isError === true
```
kayraucklnc, line 745: For the record: **this one is correct.** The TS mcp client returns a plain parsed object that still carries `isError`, so this holds. I verified it end-to-end against a real `Server`/`Client` pair over `InMemoryTransport` — `invoke_error` is recorded and the span closes `ERROR`.

No change needed. The Python twin is the one that needs work — noted on `telemetry.py`.

src/sdk/ts/src/catalog.ts
```diff
@@ -892,7 +893,11 @@ export function runToolInvocation<T>(
     );
     const started = Date.now();
 
-    const succeed = (_result: unknown): void => {
+    const succeed = (result: unknown): void => {
+      if (reportsFailure(result)) {
```
kayraucklnc, line 897: Non-blocking, but worth a comment in the code: for an `AsyncIterable` executor, `observeAsyncIterable` calls `onSuccess(lastValue)` with the **final yielded chunk** (`catalog.ts:984`), so this predicate inspects one chunk rather than the call outcome.

Consequence either way: a failure signalled on an earlier chunk is missed, and a final chunk that happens to carry `isError` marks the whole invocation failed. Neither direction is tested. Probably out of scope here — just worth a line so the next reader knows it was considered rather than overlooked.

kayraucklnc (changes requested):
# Section 1 — for humans

**The one-sentence version:** `isError` is a *wire-format* field, but the Python mcp client never hands you the wire format — it hands you a parsed `CallToolResult` model. So `isinstance(result, dict)` is `False` for every real MCP call and the Python half of this PR never fires. TypeScript is correct, verified end-to-end.

The design itself is right: one shared predicate so both channels agree, and reusing `invoke_error` rather than minting a new event type. Five findings, two blocking. Line-by-line detail is in the inline comments.

**1. The Python predicate never fires on a real MCP call** — *blocking*
Ran this branch against a real MCP server returning `isError`. Events came back `['invoke_start', 'upstream_invoke', 'invoke_end']` — the PR's own goal failing. Second trap: `mcp` 2.x renamed the field to `is_error`, so checking the `isError` attribute alone would fix 1.x and still miss the version CI resolves. Verified patch is in the inline comment on `telemetry.py`.

**2. Neither new test can detect that** — *blocking*
Both Python tests return a literal dict — the wire shape, which a real MCP call never produces. They aren't vacuous (I reverted each fix site and each test went red); they're bound to the wrong *input*, which is much harder to spot. `tests/test_mcp.py` already has the real-client harness for this.

**3. `upstream_invoke` still fires on a failed call**
Visible in the trace above. ADR-0020 defines `upstream_error`, but `mcp.py` only reaches for it when `call_tool` raises. Same bug one layer out. Fine as a follow-up, just worth being a deliberate split.

**4. The predicate is global, not MCP-scoped**
Mostly coherent, since ratel's own tools use the same convention. The one to think about is `needs_auth`, where the tool deliberately did not run and now counts toward the error rate this PR is trying to make trustworthy.

**5. Docs and changelog**
ADR-0020 defines `invoke_error` as "the executor threw"; this widens it to "threw or returned an MCP failure". Neither CHANGELOG got an `[Unreleased]` entry — #152, #154, #143 and #147 all did.

---

# Section 2 — copyable prompt for an AI agent

**Machine-oriented.** Copy the code block below into your AI coding agent — it is self-contained, with the context, the fix, the test, and the verification steps. Nothing above needs to be pasted with it.

````text
In the ratel repo on branch fix/tool-failure-telemetry, fix a bug in PR #160.

## Context

MCP signals a failed tool call by returning a successful protocol response whose
body carries `isError: true`, rather than by raising. PR #160 added a shared
predicate (`reports_failure` / `reportsFailure`) so both telemetry channels (the
OTel `execute_tool` span and the local runtime-event stream) record that as a
failure instead of a success.

The TypeScript half is CORRECT — verified against a real Server/Client pair over
InMemoryTransport. Do not change TypeScript behaviour.

## The bug (Python only)

`src/sdk/python/ratel_ai/telemetry.py` defines:

    def reports_failure(result: Any) -> bool:
        return isinstance(result, dict) and result.get("isError") is True

But `src/sdk/python/ratel_ai/mcp.py` wires each executor to
`await session.call_tool(...)`, and the mcp Python client returns a pydantic
`CallToolResult` model, never a dict. So the predicate is False for every real
MCP result, failed or not.

Second trap: mcp 2.x renamed the model field to `is_error`, keeping `isError`
only as a serialization alias. An attribute check on `isError` alone fixes 1.x
and still misses 2.x, which is what CI currently resolves.

  mcp 1.29.1 -> CallToolResult, field `isError`
  mcp 2.1.1  -> CallToolResult, field `is_error` (alias `isError`)

## Task 1 — fix the predicate

In `src/sdk/python/ratel_ai/telemetry.py`, replace the body of `reports_failure`
so it handles BOTH the dict shape (ratel's own tools return dicts) and the
pydantic model shape, and BOTH field spellings:

    def reports_failure(result: Any) -> bool:
        """MCP reports a failed call in the result (`isError: True`), not by raising.

        The mcp client returns a `CallToolResult` model, not a dict, and 2.x renamed
        the field to `is_error` (alias `isError`), so cover both shapes and spellings.
        """
        if isinstance(result, dict):
            return result.get("isError") is True or result.get("is_error") is True
        return getattr(result, "is_error", None) is True or getattr(result, "isError", None) is True

This exact patch has been verified: it flips the real-MCP probe to invoke_error,
keeps the full suite green (333 passed), and is ruff- and mypy-clean on both
mcp 1.29.1 and 2.1.1.

## Task 2 — add the regression test that would have caught this

The two existing new tests (tests/test_catalog.py and tests/test_telemetry.py)
register a hand-written executor returning a literal dict. That is the wire
shape, which a real MCP call never produces, so they cannot catch this class of
bug. Leave them, but ADD a test in `src/sdk/python/tests/test_mcp.py` that uses
the REAL client path.

That file already has the harness you need:
  - `_memory_client_session(server)` — async ctx manager yielding a live
    ClientSession over in-memory transport, handling both mcp 1.x and 2.x
  - `_mcp_major_version()` — for the 1.x/2.x server-construction split
  - `_paginated_probe_server()` — a worked example of that version split

Write a probe server with one tool whose call handler returns a CallToolResult
with isError set. Register it with `register_mcp_server(catalog, name="demo",
session=session)`, drain the trace events, `await catalog.invoke("demo__boom",
{})`, and assert the emitted event types contain "invoke_error" and NOT
"invoke_end".

Follow the existing file's conventions exactly: it uses
`from __future__ import annotations`, and any mcp handler annotated with a
`types.` name needs that module resolvable at MODULE scope for get_type_hints
(see PR #161 — do not reintroduce that bug).

## Task 3 — verify properly

Do NOT stop at "tests pass". Prove the fix is load-bearing:

  cd src/sdk/python
  uv venv --python 3.11 .venv
  uv pip install --python .venv maturin pytest pytest-asyncio ruff mypy mcp \
    opentelemetry-sdk -e ../../telemetry/python
  .venv/bin/maturin develop
  .venv/bin/ruff check . && .venv/bin/mypy ratel_ai && .venv/bin/pytest

Then, critically:
  1. Revert ONLY the telemetry.py predicate change and confirm your NEW
     test_mcp.py test goes RED. If it stays green it is not testing the fix.
  2. Re-run the suite against mcp 1.x as well:
       uv pip install --python .venv 'mcp<2' && .venv/bin/pytest
       uv pip install --python .venv 'mcp>=2' && .venv/bin/pytest
     Both must pass.

## Out of scope — do NOT do these in this change

Leave these alone; they are separate follow-ups:
  - `upstream_invoke` vs `upstream_error` in mcp.py
  - scoping the predicate to MCP-only tools
  - ADR-0020 / CHANGELOG updates
````

src/sdk/python/ratel_ai/telemetry.py
faustoq: Fixed in [a233f3b](https://github.com/ratel-ai/ratel/pull/160/changes/a233f3b575ef10ff7f4e898a28ecf7c081360bdd): both shapes, both spellings. Your point about 2.x holds: the attribute lands as `isError` on 1.29.1 and `is_error` on 2.1.1, so an "isError - only" check would have passed locally and missed CI.

src/sdk/python/ratel_ai/catalog.py
faustoq: This branch is now reachable. The new `test_mcp.py` test drives a real `ClientSession` through this exact path and gets `invoke_error`.

src/sdk/python/tests/test_catalog.py
faustoq: Kept, and added the real-client test in `tests/test_mcp.py` on the `_memory_client_session` harness. 

Reverting the predicate turns it red with exactly your trace: `assert 'invoke_error' in ['invoke_start', 'upstream_invoke', 'invoke_end']`.

src/sdk/python/tests/test_telemetry.py
faustoq: the span now closes `ERROR` on a `CallToolResult`. 

It feeds the model directly rather than through the live client, since the global OTel provider registered here can't be re-registered from `test_mcp.py`

src/sdk/ts/src/telemetry.ts
faustoq: Thanks for verifying it end-to-end! Left untouched

src/sdk/ts/src/catalog.ts
faustoq: Confirmed: `onSuccess(lastValue)` means the predicate sees the final chunk, not the call outcome.

Leaving it to the follow-up that decides what a streaming failure should mean.

kayraucklnc (approved):
Both blockers verified fixed.

PR #161, opened 2026-08-31
fix(sdk): resolve handler annotations in the paginated MCP test

faustoq:
## Problem

Test `tests/test_mcp.py::test_register_mcp_server_paginated_list_tools_real_client_session` fails against mcp 1.28 with the following error: 

```
handle_list_tools() missing 1 required positional argument: 'request'
```

The file uses `from __future__ import annotations`, so annotations are strings evaluated at runtime. The 1.x handler is annotated with `types`, which is imported inside `_paginated_probe_server` and is therefore not a module global. `get_type_hints` raises `NameError`, and mcp's `create_call_wrapper` catches that and falls back to the old handler style, calling the handler with no arguments.

`pyproject.toml` pins `mcp>=1.0` with no upper bound, so a newer mcp introduced the two-style handler dispatch that exposes this.

## Solution

Import the module at module scope so `get_type_hints` can resolve the annotations. The handler signature is unchanged. The import is guarded because the suite also runs without mcp installed.

Only `list_tools` is affected: `call_tool` does not go through `create_call_wrapper`, and the TypeScript SDK registers handlers with `setRequestHandler`, which always passes the request.

Added CI step to run the tests using `mcp 1.x`, to ensure retro-compatibility in the future.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -417,10 +417,12 @@ jobs:
       # `mcp` is installed unpinned and currently resolves to 2.x, so the 1.x-only
       # code path in `_paginated_probe_server` is never hit. `pyproject.toml` allows
       # mcp>=1.0; re-run the suite on 1.x to keep that retro-compatible path verified.
+      # Then restore the resolved version; later steps share this venv.
       - name: Test backwards compatibility with mcp 1.x
         run: |
           uv pip install --python .venv 'mcp<2'
           .venv/bin/pytest
+          uv pip install --python .venv 'mcp>=2'
       # Python 3.9 floor (requires-python >=3.9; ADR-0006 ships cp39 abi3 wheels). The
       # suite above runs on 3.11 only, and the AST guard (tests/test_py39_compat.py)
       # catches just the PEP-604 class of import breakage — stdlib/typing drift (e.g.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -414,6 +414,13 @@ jobs:
         run: .venv/bin/mypy ratel_ai
       - name: Test
         run: .venv/bin/pytest
+      # `mcp` is installed unpinned and currently resolves to 2.x, so the 1.x-only
+      # code path in `_paginated_probe_server` is never hit. `pyproject.toml` allows
+      # mcp>=1.0; re-run the suite on 1.x to keep that retro-compatible path verified.
+      - name: Test backwards compatibility with mcp 1.x
+        run: |
+          uv pip install --python .venv 'mcp<2'
+          .venv/bin/pytest
       # Python 3.9 floor (requires-python >=3.9; ADR-0006 ships cp39 abi3 wheels). The
       # suite above runs on 3.11 only, and the AST guard (tests/test_py39_compat.py)
       # catches just the PEP-604 class of import breakage — stdlib/typing drift (e.g.
--- a/src/sdk/python/tests/test_mcp.py
+++ b/src/sdk/python/tests/test_mcp.py
@@ -16,6 +16,11 @@ from typing import Any
 
 import pytest
 
+try:  # module scope so get_type_hints can resolve the handler annotations below
+    from mcp import types
+except ImportError:
+    types = None  # type: ignore[assignment]
+
 from ratel_ai import (
     EmbedderError,
     McpToolsListError,
--- a/src/sdk/python/tests/test_mcp.py
+++ b/src/sdk/python/tests/test_mcp.py
@@ -79,7 +79,6 @@ async def _memory_client_session(server: Any) -> AsyncGenerator[Any, None]:
 def _paginated_probe_server() -> Any:
     """Server that returns two pages of tools/list (decorator API on 1.x, handlers on 2.x)."""
     if _mcp_major_version() >= 2:
-        import mcp_types as types
         from mcp.server.lowlevel.server import Server
 
         tools = [
@@ -111,7 +110,6 @@ def _paginated_probe_server() -> Any:
             on_call_tool=on_call_tool,
         )
 
-    from mcp import types
     from mcp.server import Server
 
     server = Server("ratel-pagination-probe")


kayraucklnc (commented):
## Review summary

Checked this out into a worktree and ran it on both `mcp` lines. **This is a good PR** — correct diagnosis, minimal fix, and the CI step you added is genuinely load-bearing rather than ceremonial. Two optional nits below, neither blocking.

What I verified:

| Check | Result |
|---|---|
| Reproduce the reported failure on `mcp` 1.29.1, without your fix | reproduced exactly — `handle_list_tools() missing 1 required positional argument: 'request'` |
| `tests/test_mcp.py` with your fix, `mcp` 1.29.1 | 13 passed |
| `tests/test_mcp.py` with your fix, `mcp` 2.1.1 | 13 passed |
| `ruff check .` + `mypy ratel_ai` | clean |

> **How to read this review.** The first section is written **for a human**. The second is a **copyable prompt for an AI coding agent** covering the two nits. Both are optional here — the PR is correct as it stands.

---

# Section 1 — for humans

### Why this fix is the right one

Your write-up nails it, and I want to restate the mechanism because it's a genuinely nasty failure mode worth having written down somewhere.

`from __future__ import annotations` turns every annotation into a **string** that gets evaluated later, by whoever needs it. mcp's `create_call_wrapper` needs it: it calls `get_type_hints()` to decide whether a handler is the new request-taking style or the old zero-argument style. `get_type_hints()` evaluates those strings against the function's **module globals** — and `types` was imported inside `_paginated_probe_server`, so it's a local. `NameError`.

The part that makes it painful is what happens next: mcp **catches** that `NameError` and quietly falls back to the old handler style. There's no error about an unresolvable annotation — just a handler invoked with the wrong arity, several layers away from the actual cause. Moving the import to module scope is exactly the right fix, and leaving the handler signature untouched is the right instinct.

Your scoping note checks out too: only `list_tools` goes through `create_call_wrapper`, and the TS SDK uses `setRequestHandler`, which always passes the request.

### The CI step is the valuable part

I want to call this out specifically because it would be easy to read as boilerplate. `pyproject.toml` pins `mcp>=1.0` with no upper bound, CI resolves to 2.x, so **without your second leg the entire 1.x branch of `_paginated_probe_server` is never executed at all**. I confirmed it catches the real thing by reproducing the failure it guards. Good instinct.

### Nit 1 — the `mcp<2` downgrade is sticky for the rest of the job

`uv pip install --python .venv 'mcp<2'` mutates the shared venv and nothing puts it back. Harmless *today*: the only later step touching that venv is `maturin build`, and the 3.9 leg and the pydantic-ai example both build their own environments. But a step added below it later would silently test against 1.x, and that would be a confusing afternoon. One line to remove the trap.

### Nit 2 — the local import that caused the bug is still there

`from mcp import types` at `test_mcp.py:114` is now redundant, and it's precisely the import that does **not** resolve the annotation. Leaving it means the shape of the original bug survives in the file, and someone tidying up the module-level import later reintroduces it. Your comment on the new import mitigates this well, so it's genuinely minor — but deleting the local one would close the loop.

(Related and harmless: on 2.x the module global is `mcp.types` while the 2.x branch locally aliases `mcp_types`. Different module objects, same classes — `mcp.types.CallToolResult is mcp_types.CallToolResult` is `True` — so nothing breaks.)

### One thing I'd leave exactly as you wrote it

Re-running the whole suite on 1.x costs ~35s. Scoping it to `tests/test_mcp.py` would be cheaper, but running everything also catches unrelated 1.x drift, which is arguably the more valuable guard. Keep it.

---

# Section 2 — copyable prompt for an AI agent

**This section is machine-oriented.** Copy the code block into an AI coding agent if you want the two nits applied. Both are optional — nothing here is blocking.

````text
In the ratel repo on branch fix/mcp-test-annotations, apply two small cleanups
from review of PR #161. Neither is a bug fix; the PR is already correct.

## Cleanup 1 — don't leave the CI venv downgraded

In `.github/workflows/ci.yml`, the step "Test backwards compatibility with mcp 1.x"
runs `uv pip install --python .venv 'mcp<2'` and never restores the venv, so every
later step in that job sees mcp 1.x. Nothing after it depends on 2.x today, but a
future step would silently test the wrong version.

Fix it so the downgrade cannot leak. Either restore afterwards:

    - name: Test backwards compatibility with mcp 1.x
      run: |
        uv pip install --python .venv 'mcp<2'
        .venv/bin/pytest
        uv pip install --python .venv 'mcp>=2'

or run the 1.x leg in a throwaway venv. Prefer whichever matches the conventions
already used elsewhere in this workflow file. Keep the existing explanatory
comment above the step, and keep running the FULL suite (not just test_mcp.py) —
that breadth is deliberate.

## Cleanup 2 — remove the now-redundant local import

In `src/sdk/python/tests/test_mcp.py`, `_paginated_probe_server` still does
`from mcp import types` inside the function (around line 114), in the mcp 1.x
branch. The module-scope import added by this PR is what actually resolves the
handler annotations via get_type_hints; this local one does not, and is exactly
the construct that caused the original bug.

Delete that local import line so the 1.x branch uses the module-level `types`.
Do NOT touch:
  - the module-scope guarded import at the top of the file
  - its explanatory comment
  - the handler signatures
  - the `import mcp_types as types` in the 2.x branch

## Verify

  cd src/sdk/python
  .venv/bin/ruff check . && .venv/bin/mypy ratel_ai

  # must pass on BOTH mcp lines
  uv pip install --python .venv 'mcp<2'  && .venv/bin/pytest tests/test_mcp.py
  uv pip install --python .venv 'mcp>=2' && .venv/bin/pytest tests/test_mcp.py

Then confirm cleanup 2 did not silently break annotation resolution: with mcp 1.x
installed, `test_register_mcp_server_paginated_list_tools_real_client_session`
must PASS. If it fails with "missing 1 required positional argument: 'request'",
you removed the wrong import.
````

---

*Reviewed against `main` @ `cf3e9c2`. No objections from me — I've left this as a comment rather than a formal approval so Kayra can make that call.*

File: .github/workflows/ci.yml [hunk 1]
```diff
@@ -414,6 +414,13 @@ jobs:
         run: .venv/bin/mypy ratel_ai
       - name: Test
         run: .venv/bin/pytest
+      # `mcp` is installed unpinned and currently resolves to 2.x, so the 1.x-only
+      # code path in `_paginated_probe_server` is never hit. `pyproject.toml` allows
+      # mcp>=1.0; re-run the suite on 1.x to keep that retro-compatible path verified.
+      - name: Test backwards compatibility with mcp 1.x
+        run: |
+          uv pip install --python .venv 'mcp<2'
```
kayraucklnc, line 423: Nit: this downgrade is **sticky** — it mutates the shared `.venv` and nothing restores it, so every later step in this job runs against mcp 1.x.

Harmless today (only `maturin build` uses this venv afterwards; the 3.9 leg and the pydantic-ai example build their own), but a step added below this one would silently test the wrong version.

Either append `uv pip install --python .venv 'mcp>=2'` after the pytest run, or do the 1.x leg in a throwaway venv.

To be clear though — the step itself is a good call. Without it the entire 1.x branch of `_paginated_probe_server` is never executed, since CI resolves mcp to 2.x. I confirmed it catches the real regression.

src/sdk/python/tests/test_mcp.py
```diff
@@ -16,6 +16,11 @@
 
 import pytest
 
+try:  # module scope so get_type_hints can resolve the handler annotations below
+    from mcp import types
```
kayraucklnc, line 20: This is the right fix, and the comment explaining *why* it must be at module scope is the thing that makes it survive future cleanup. 

One follow-on: `from mcp import types` inside `_paginated_probe_server` (~line 114, not in this diff) is now redundant, and it's precisely the import that does **not** resolve the annotations. Leaving it means the shape of the original bug is still in the file, and anyone tidying up this module-level import later reintroduces the failure. Worth deleting the local one to close the loop.

.github/workflows/ci.yml
faustoq: thanks for catching this! fixed

src/sdk/python/tests/test_mcp.py
faustoq: Good catch, it should go, but line 114 isn't the redundant one and deleting it alone gives `UnboundLocalError`.

`import mcp_types as types` (line 82) binds `types`, and Python scoping is static, so the name is local to the whole function, including the 1.x branch, which never runs line 82. Line 114 is that path's only binding.

Line 82 is the real redundancy: on 2.x `mcp.types` re-exports the identical classes, so the alias buys nothing and creates the shadow.

Deleted both instead: the module-scope import is now the single binding. 
Green on mcp 1.x and 2.x. Removing the module-scope import still turns the test red, so it stays load-bearing.

kayraucklnc (approved):
Both nits addressed.

PR #163, opened 2026-09-07
fix(core): key adaptive-ranking pending state by turn_id

bercaakbayir:
## Summary

Two agent sessions sharing one catalog/graph could cross-pair a search
from one with an invoke from another: `UsageLearner`'s pending search and
`IntentGraph`'s `PendingQuery`/`CreditSlot` were single unkeyed slots.
Session A searches X, session B searches Y and clobbers the slot, session A
invokes Z — the graph learns Y → Z instead of X → Z.

This isn't a newly-discovered bug: ADR-0014 already named it and deferred
the fix ("a per-turn correlation id threaded through the trace events,
deferred as not worth the plumbing"). This PR implements that deferred
fix.

- `TraceEventContext`/`TraceEnvelope` gain an optional `turn_id`, distinct
  from the trace-*stream* `session_id` fixed at sink construction.
- `UsageLearner.pending` and the graph's `PendingQuery`/`CreditSlot` are now
  keyed by `turn_id` (bounded, FIFO-evicted past a cap), with a reserved
  sentinel key reproducing today's single-slot behavior — cross-session
  collisions included — for callers that supply none. Fully additive: no
  existing caller's behavior changes.
- TS and Python SDKs expose an optional `turnId`/`turn_id` on
  `search`/`searchAsync`/`invoke`/`invokeRaw` (tool and skill catalogs),
  threaded through the existing projection/context plumbing.
- Amends ADR-0014 in place to record the fix.

## Test plan

- [x] New Rust tests reproduce the cross-pairing bug (fail pre-fix, pass
      post-fix): `two_concurrent_sessions_with_turn_ids_do_not_cross_pair`,
      plus the graph-level `CreditSlot`/`PendingQuery` equivalents.
- [x] `no_turn_id_reproduces_todays_single_slot_pairing_exactly` pins the
      backward-compat fallback.
- [x] All pre-existing tests pass unmodified (397 Rust, 632 TS, 333 Python).
- [x] New end-to-end tests in both SDKs exercise `turnId` through the real
      native binding.
- [x] `cargo test/clippy/fmt`, TS `typecheck/lint/test`, Python
      `ruff/mypy/pytest` all green.


rstagi (commented):
## Verdict

4 majors, no blockers. Right idea, careful implementation, but the fix stops short of the topology it targets and one ADR claim this PR writes is false in the code it ships. All four inline.

Only majors here. I also found 9 lower-severity items (test gaps the green suite hides, unamended ADR-0020 envelope table, a stale comment, a `context?: unknown` param-order footgun, copy-paste)

src/core/src/usage_learner.rs
```diff
@@ -238,13 +246,17 @@ pub(crate) fn replay_log_into(
         // it an observation; later ones add edges for the same question.
         let first_confirmation = !entry.1;
         entry.1 = true;
-        // Stash this query's vector right before the observation reads it. The
-        // slot holds one entry, so with sessions interleaved anything set
-        // earlier may belong to another session's question.
+        // Stash this query's vector right before the observation reads it,
+        // under the shared NO_TURN key: replay walks the log sequentially, one
```
rstagi: **[major] replay ignores `turn_id`, so offline graphs cross-pair the turns the live path separates**

This comment's conclusion is wrong. `replay_log_into` keys pending by `env.session_id` and never reads `env.turn_id`. But `session_id` is fixed once per sink (`trace/sink.rs:62`), so in the exact deployment this PR targets (concurrent sessions, one catalog, one sink, each with its own `turn_id`) every envelope shares one `session_id` and differs only by `turn_id`. Sequential walking protects the vector stash, not the pairing.

Same 4-event log, both paths:

```
LIVE:    [("delete a stale branch", ["git_branch_delete"]), ("rotate the signing key", ["vault_rotate"])]
OFFLINE: [("rotate the signing key", ["git_branch_delete", "vault_rotate"])]
```

Turn A's query vanishes, its invoke lands on turn B's question: the pre-PR bug, reintroduced in `build_intent_graph` (`tool_registry.rs:372`). Seeded evidence, so the wrong edge boosts the wrong tool permanently.

This contradicts the amendment this PR writes. `docs/adr/0014:276` claims "A live caller that supplies `turn_id` does not diverge here", and the bullet above says equivalence is asserted by test. Every equivalence and interleaving test feeds `turn_id: None` envelopes, so nothing catches it.

Fix: key replay by `(session_id, turn_id)`, falling back to `session_id` when absent. Preserves existing tests and logs.

Prevention: the equivalence test is the intended guard and it stopped covering the live keying the moment that key changed. Parameterize it over turn-keyed logs, not just `turn_id: None`.

Also: `docs/baseline-capture-distributed.md:57` still tells users to build a sink per turn or restamp `session_id`. That workaround is still load-bearing because of this gap, and `turn_id` was meant to retire it.

src/sdk/ts/src/catalog.ts
```diff
@@ -505,9 +509,16 @@ export class ToolCatalog {
     topK: number,
     origin: SearchOrigin = "direct",
     method?: SearchMethod,
+    turnId?: string,
```
rstagi, line 512: **[major] no `turnId` path through capability tools or the `ratel()` facade, in either SDK**

This doc says to pass the same id to `invoke` "when multiple concurrent sessions share this catalog's graph", but the surfaces where that actually happens got no plumbing:

- TS: `capabilities.ts:557-558` (`invokeValidatedRaw`/`invokeRaw`), `:355` and `:405` (`searchAsync`), `skill-tools.ts:70`, facade handles at `ratel.ts:595-598`, `:616`, `:663`, `:716`.
- Python: same twins at `capabilities.py:171`, `:225`, `:394`, `skill_tools.py:58`.

`grep -n turnId` over those TS files: zero hits. No option or context carries a turn id, so a host on standard `modelTools()` wiring cannot opt in at all.

Partial adoption is worse than none, and silent. Pre-rank with `searchAsync(q, 5, "direct", undefined, "turn-7")`, let the model execute via `invoke_tool`: the search arms under `turn-7`, the invoke confirms under `NO_TURN`. `confirm` then either finds nothing (all learning dropped, forever, while the docs say you opted in) or finds a stale `NO_TURN` query and credits another session's question to this session's tool: the cross-pair this PR exists to close.

Either thread `turnId` through the funnel and facade, or scope the docs and ADR to direct-catalog callers and call the funnel a follow-up.

src/sdk/ts/src/telemetry.ts
```diff
@@ -948,9 +963,10 @@ export function traceSearchAsync<T extends { length: number }>(
 export function traceSkillLoad<T>(
   skillId: string,
   run: (projection: RuntimeEventProjection) => T,
+  turnId?: string,
```
rstagi, line 966: **[major] this `turnId` is dead in TS, and the resulting asymmetry double-counts support**

Nothing in TS supplies it. `skill-catalog.ts` is untouched: `search` (`:273`), `searchAsync` (`:285`), `invoke` (`:414`, calling `traceSkillLoad` with two args). Python's `SkillCatalog` has `turn_id` on all three, so the SDKs diverge, and the PR description claims both ("tool and skill catalogs").

Worse than a missing surface: in the tool+skill fan-out `CreditSlot` exists to dedupe, the tool side can key and the skill side cannot, so dedup breaks. One question, tool side `turn-1`, skill side forced to `NO_TURN`:

```
assertion `left == right` failed: one fanned-out question should still credit once
  left: 2
 right: 1
```

`support` scales the boost and gates the flip, so a partially-keyed TS host inflates the number ADR-0014 is most careful about. Note the direction: keying only the tool side makes support worse than not opting in, which an additive flag should not do.

Adding `turnId` to the three `skill-catalog.ts` methods closes it and gives this param a caller.

src/core/src/usage.rs
```diff
@@ -89,6 +89,68 @@ const EVICTION_FLOOR: f32 = 0.01;
 /// memory; the centroid is a running mean and is unaffected by dropping members.
 const MEMBER_CAP: usize = 50;
 
+/// Reserved pending-state key for callers that never supply a `turn_id` on
+/// [`crate::trace::TraceEventContext`]. Every such caller shares this one
+/// slot, reproducing the single-slot pairing behavior that predates `turn_id`
+/// — including its cross-session collisions, which is the documented,
+/// accepted cost of opting out (ADR-0014).
+pub(crate) const NO_TURN: &str = "";
```
rstagi, line 97: **[major] a caller-supplied empty `turn_id` silently becomes the shared legacy slot**

`NO_TURN` is `""` and every site resolves `turn_id.as_deref().unwrap_or(NO_TURN)`, so `Some("")` and `None` are indistinguishable. Empty strings are not exotic: `session.id ?? ""`, an empty env var, `String::new()`.

Caller A opts in with `Some("")`, caller B never opted in, interleaved, A invokes:

```
assertion `left == right` failed: an explicit turn_id must not share caller B's un-keyed slot
  left: Some("rotate the signing key")
 right: Some("delete a stale branch")
```

A caller that believes it is protected gets the exact corrupt edge this PR prevents, with no error and no trace of it. Immediate fix is one line: treat empty as absent (or reject it).

Prevention: the opt-out sentinel is a valid value of the caller-supplied type, so any caller can collide with it. Key by `Option<&str>`, or use a sentinel no caller can spell, and the class is gone rather than this instance.

src/core/src/usage_learner.rs
bercaakbayir: (086c3e5)  pending state is now keyed by (session_id, turn_id) instead of session_id alone. A log with no turn_id set falls back to exactly the old per-session behavior, so nothing existing changes.

src/sdk/ts/src/catalog.ts
bercaakbayir: (6cc2bb7) threaded turnId/turn_id through the full funnel in both SDKs: searchCapabilitiesTool/invokeToolTool/getSkillContentTool (and their Python equivalents), runCapabilitiesSearch, and ratel.ts's ToolCollection/AdaptedToolCollection/recall.

src/sdk/ts/src/telemetry.ts
bercaakbayir: (6cc2bb7) added turnId?: string to SkillCatalog.search/searchAsync/invoke, bringing it to parity with ToolCatalog.

src/core/src/usage.rs
bercaakbayir: (19e6440)  BoundedMap and everything downstream of it now key by Option<&str> instead of &str, so "no turn_id" is just None, not a string that happens to be "".

rstagi (approved):

